Issue #253, opened 2014-05-27
Given object, make roxygen template


hadley:
Maybe `roxygen_template()`. Could be used for rstudio and ess.

Should add `@export` and `@examples` by default. Add `@return` if it's a function (incl. generics)

ateucher:
@hadley I don't know if it's helpful or along the lines you're thinking, but I've written a function that I use a lot for this: https://gist.github.com/ateucher/1342ec5c7f2ab3d83c32.

hadley:
Thanks @ateucher - I think that provides a bit too much (since a lot of it will need to be deleted after generating it), but it's helpful.

ateucher:
Gotcha. True, I do end up deleting a lot of it...

hadley:
cc @kevinushey

kevinushey:
FWIW the current RStudio implementation is done through a combination of R and GWT code: https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/common/r/roxygen/RoxygenHelper.java; I'm guessing it's not that helpful for `roxygen2`.

I think the simplest mechanism for RStudio would be something of the form:

1. We given `roxygen2` the current function, as well as its roxygen comment, as a string,
2. `roxygen2` returns us text that we use to replace the current roxygen2 block (or generate the skeleton if none is there).

hadley:
@kevinushey has anything changed here? I'm guessing implementing this in roxygen and getting merged into the IDE would be relatively expensive.

hadley:
Given that RStudio now does this for you, I'm going to close this (old) issue.

hechth:
Will this feature be supported in other editors? I'm specifically looking at VSCode

PR #263, opened 2014-07-15
Add a (small) function to generate params list

sahilseth:
This should help in making templates for params. Could expand this to generate a template with 
@params @title, @values, @examples etc..


hadley:
I'm going to close for now. See #253 for more discussion.

PR #287, opened 2014-09-30
Use %==% to check for identical equality

jimhester:
This improves safety and correctness over using == for testing 1 to 1
equality.

As you probably know, `==` has issues when the arguments are not as you were expecting.
i.e.
```r
if(names(1) == 'val') { print('val') }
# Error in if (names(1) == "val") { : argument is of length zero

if(c(1,2) == 'val') { print('val') }
#Warning message:
#In if (c(1, 2) == "val") { :
#  the condition has length > 1 and only the first element will be used
#logical(0)
```

Using `%==%`, which is just an alias for `identical()` does the correct thing in these cases while maintaining readability of the expression.

```r
if(names(1) %==% 'val') { print('val') }
if(c(1,2) %==% 'val') { print('val') }
```

I thought `roxygen` could use some more robustness in this regard.  So I went through and replaced all the one to one comparisons that were previously using `==` with `%==%`.  All the tests pass on my machine, so there should not be any user visible change.


hadley:
Could you give a little motivation for the change?

jimhester:
I don't have a specific use case where the current behavior fails.

I just find defining `%==%` useful because while there are 'longer form' operators `&&` and `||` for control flow there is no equivalent operator for equality. Only the element-wise `==`, which is equivalent to `&` and `|` (with the same inherit drawbacks).

I am fine closing this pull if you don't find it useful.

hadley:
It seems useful, but roxygen doesn't feel like the right place for it. Maybe purrr? But that doesn't feel quite right either.

PR #312, opened 2014-12-10
Merge sections with identical title

### LINKED ISSUE #300: Merge sections with identical title
If two files have entries `@section OBTW` and `@rdname base`, the file `base.Rd` contains two sections titled "OBTW". On the other hand, the "Description" and "Detail" sections are merged. Wouldn't it be a good idea (and feasible) to also merge sections with identical names?

Example package: https://github.com/krlmlr/section.merge.test

Also, the contents are generated in the order of the input file names. In the example above, it's a < b < base, but `base.R` is the "main" source file for `base.Rd`. Perhaps the contents that come from the "main" input file (if there is one) should come before the other contents?

krlmlr:
- Add test
- In order of appearence (it would be better to use the contents of the "main" file first)

Fixes #300.


R/rd-tag-api.R
```diff
@@ -53,6 +53,29 @@ merge.minidesc_tag <- function(x, y, ...) {
   list(x)
 }
 
+#' @export
+merge.section_tag <- function(x, y, ...) {
+  x_names <- sapply(x$values, `[[`, "name")
```
hadley: Could you please use `vapply()` here?

hadley:
Could you please also update news and squash?

krlmlr:
Now using vapply() instead of sapply() (there are still two instances of sapply() in the code, in process_def_tag()). Moved and changed NEWS.

Why are squashed pull requests important to you? I've only recently discovered `git log --first-parent`, which does the "right thing" for the master branch. See also http://devblog.nestoria.com/post/98892582763/maintaining-a-consistent-linear-history-for-git.

hadley:
It's not super important, but there are 16 commits here for a relatively small change - it makes understanding the history that much harder if I look at it in the future

krlmlr:
The history is an issue, but this is what `--first-parent` is for. Merged pull requests are shown as a single unit, without the detailed commits. (Also works with `gitk`, but then I don't know which tool you use to browse history.)

I'll squash the commits in a minute.

krlmlr:
Removed and squashed irrelevant commits.

hadley:
Thanks!

PR #325, opened 2015-02-25
Only update changed collate

Geoff99:
Hi @hadley 

This pull request relates to devtools issue hadley/devtools#723

@renozao neatened up my first quick fix to make `update_collate` rewrite the Collate entry in the DESCRIPTION file *only* when roxgyen2 finds changes generated by `#' @includes` in the package.  He also wrote a unit test as well.  

This code won't guarantee that devtools can run safely *in parallel* (that's a much bigger job I think) but I think it will resolve the use case which generated the devtools issue (or at least return control of the problem to  anyone running multiple `load_all`s at the same time on a package)

Happy to respond to any changes you'd like made. Thanks in advance for looking over this.


renozao:
I don't think we need to tackle the parallel loading in general.

I believe that for plain R packages it should be enough to ensure that no files is changed if not necessary. The use case is that the package has first been loaded by ``load_all`` in a separate session **before** running the parallel jobs, so that each job, when individually loading the package with ``load_all`` should not overwrite anything.

hadley:
Could you please add a bullet point to NEWS?

Geoff99:
Bullet point about `update_collate` only rewriting the Collate entry after changes has been added to NEWS, as requested.

hadley:
Can you please put the NEWS under the top-most bullet point? (should be under 4.1.1.9000) (and similarly for your other PR)

Geoff99:
NEWS entries moved.

hadley:
Thanks!

PR #328, opened 2015-03-19
Clarify overwrite prevention as reason for skipping file

coolbutuseless:
Text change to explain that an existing file which was not created by roxygen is not being overwritten.

Keeping it within 80cols makes it look a little awkward.


hadley:
Thanks!

PR #330, opened 2015-03-29
Keep whitespace in Authors@R field

jranke:
As the structure of the Authors@R field is potentially a bit more
complex than that of the other fields, this commmit leads to whitespace
being preserved when reading in the Authors@R field.

### DIFF
--- a/tests/testthat/test-wrapFieldIfNecessary.R
+++ b/tests/testthat/test-wrapFieldIfNecessary.R
@@ -56,10 +56,26 @@ test_that("DESCRIPTION fields DO NOT get wrapped if no line exceeds the wrapping
 )
 
 test_that("Infinity threshold turns off wrapping", {
-  poem <- paste(sample(letters, 1000, TRUE), collapse = "")
-  expect_equal(
-    wrap_field_if_necessary("LongPoem", poem, wrap.threshold = Inf),
-    paste("LongPoem:", poem)
-  )
-}
+    poem <- paste(sample(letters, 1000, TRUE), collapse = "")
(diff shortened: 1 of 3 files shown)


hadley:
Could you add a test for this? It's a bit tricky because you might also need to preserve on output too - or maybe roxygen should be preserving whitespace for all inputs?

tests/testthat/test-wrapFieldIfNecessary.R
```diff
@@ -56,10 +56,26 @@ test_that("DESCRIPTION fields DO NOT get wrapped if no line exceeds the wrapping
 )
 
 test_that("Infinity threshold turns off wrapping", {
-  poem <- paste(sample(letters, 1000, TRUE), collapse = "")
-  expect_equal(
-    wrap_field_if_necessary("LongPoem", poem, wrap.threshold = Inf),
-    paste("LongPoem:", poem)
-  )
-}
+    poem <- paste(sample(letters, 1000, TRUE), collapse = "")
```
jranke: This is just a formatting change

jranke:
OK, so a test has been added to the pull request that checks that the 
formatting of the Authors@R field is preserved (which it wasn't in my original 
commit...). As far as the other fields are concerned, the present way they are 
reformatted is fine with me.

Thanks for maintainin roxygen2.

Johannes

Am Montag, 30. M�rz 2015, 07:15:02 schrieb Hadley Wickham:

hadley:
Oops, forgot to ask for a sentence in the news too. Can you please create a new PR?

PR #333, opened 2015-04-03
Trim possible white spaces around @include tags

yihui:


hadley:
I wonder if this should also fail with an obvious message if the file doesn't exist?

yihui:
Maybe. I guess you can add such a message pretty quickly after merging this PR :)

hadley:
Thanks!

PR #335, opened 2015-04-27
render Rd without roxygen2 version

leeper:
This stems from [a post on StackOverflow](http://stackoverflow.com/questions/29711413/r-render-rd-using-roxygen2-without-roxygen2-version) where someone makes the reasonable argument that changing every Rd file due only to a change in the roxygen2 package version is disruptive to version control.

An alternative approach would be to not update Rd files that have not be changed (thus allowing .Rd files from multiple versions of roxygen2 to exist in a package directory simultaneously), but I'm not familiar enough with the roxygen2 code to make that kind of change.


leeper:
I see this is already being considered (https://github.com/klutometis/roxygen/pull/334), so I'll close this.

PR #348, opened 2015-06-02
Revert "Don't write blank namespace. Closes #276"

### LINKED ISSUE #276: Roxygen2 shouldn't override namespace if no namespace tags
Need to check this happens. Otherwise the default NAMESPACE will get overwritten after you've learned about function documentation, but not namespaces.

krlmlr:
This reverts commit cc6b8aa2964cd78e9ffc43af06bd050eb3623169.

The shortcut is no more necessary, as roxygen2 won't overwrite files that do not include the "magic comment". On the other hand, the shortcut prevents creation of truly empty `NAMESPACE` files.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,9 @@
   for fields parsed from the "introduction" (the text before the first tag)
   (@gaborcsardi, #370)
 
+* Empty `NAMESPACE` file is written if it is maintained by `roxygen2`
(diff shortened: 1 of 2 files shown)


hadley:
Thanks - can you please update NEWS and squash?

krlmlr:
Squashed and moved NEWS.

hadley:
Thanks!

PR #350, opened 2015-06-03
Bump version

krlmlr:
The state of eb14242 seems to be identical to the CRAN version, please create a tag.


hadley:
Did by hand. Thanks! (and sorry I forgot to do this when I released)

PR #340, opened 2015-05-27
Miscellanea around vignette roclet

krlmlr:
- Add vignette roclet to readme and to "roclet family" in docs
- Warn about unknown fields in `Roxygen` tag in `DESCRIPTION`

PR #351, opened 2015-06-03
WIP: Fix failing tests on Travis

krlmlr:
It seems that running `stri_install_icudt()` is necessary after installing the `stringi` package via `apt`.  Another option would be to disable these tests on Travis.

Reference: https://github.com/klutometis/roxygen/pull/340#issuecomment-108127298


krlmlr:
Merged `master`. The check fails without this (or a similar fix).

hadley:
It's a bummer that we need to do this

krlmlr:
We could also `skip()` on Travis, or escalate upstream (to the `r-cran-stringi` package).

hadley:
Or escalate to the travis image, if that ever gets off the ground

krlmlr:
I really like the "bring your own image" approach by wercker. @jimhester has built wercker steps for R testing, I have a relatively light base image with devtools and an easy way to build your own base images pre-loaded with the dependencies you need.

hadley:
Unfortunately wercker is an absolute no go while they still request full read and write privileges for all public and private repos. There's absolutely no reason to need that.

krlmlr:
I wasn't aware of that -- Travis is really much more cautious here.

jimhester:
https://github.com/wercker/support/issues/20 was the issue about this, but it seems to have been closed :frowning:.  Someone from wercker indicated they would consider requiring only read permissions if you did not have a private repository, but could not give a timetable for when that would be implemented.

PR #355, opened 2015-06-08
Replace "C" Travis with "R" one.

### LINKED ISSUE #353: Move travis to proper R "version"
Hi @hadley, 

Is there any reason why one should not move travis to `R` style description. Instead of having the "old" C one style. 

Just by looking at `travis.yml` I cannot think about any.

dmpe:
Hi Hadley, 

Here I have updated travis with R, and build it without problems. See here https://travis-ci.org/dmpe/roxygen/builds/65906164

Fixes  #353


hadley:
Could you please double check the stringi work around is still necessary and then squash?

dmpe:
will check

dmpe:
https://github.com/klutometis/roxygen/pull/386

Issue #283, opened 2014-09-25
make it possible to not reference all aliases with @family


thomasp85:
It seems @family uses aliases to create cross-linking between documentation. Sometimes this is not desired as it can inflate the size of the see also section if there is cross-linking between classes with a lot of accessors documented with the class.

Is it possible to provide a toggle or a new tag that only takes the name into account?

gaborcsardi:
This would be great indeed. I think it has little use to include 5-10 links that point to the same manual page....

gaborcsardi:
FWIW, it is easy to change to this behavior (obviously): https://github.com/gaborcsardi/roxygen/commit/796fc0b2b8a79241524864f6d23ac50bcdea22a2

I am not sure what the best way would be to provide a toggle.

thomasp85:
What is the status on this? Lack of time, disagreement on the feature, or something else?

hadley:
I'd be happy to review a PR

thomasp85:
Would you prefer a change in behaviour of the current @family tag or the creation of a new e.g. @smallFamily?

hadley:
Change current behaviour

PR #367, opened 2015-08-05
@family refers to name only, not all aliases, #283

gaborcsardi:
I had this PR lying around. Details were discussed at #283. 

I am actually not convinced that this is always the good behavior. Sometimes it makes sense to link to all aliases, so if you have an idea on how to make this user configurable, per family, probably, I can update this PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # roxygen2 4.1.1.9000
 
+* `@family` now cross-links each manual page only once, instread of linking
+  to all aliases (#283, #367)
+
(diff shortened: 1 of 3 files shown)


hadley:
Yeah, we need to think this one through

hadley:
Could you please add a bullet to NEWS? I think I just need to try it out on a few of my projects and see how it feels.

gaborcsardi:
OK, NEWS updated.

hadley:
Thanks! I definitely prefer the new behaviour for roxygen2 itself

PR #370, opened 2015-08-20
Call preref funcs for automatic tags in intro, closes #364

### LINKED ISSUE #364: Preref parser for auto-detected title, description and details
The registered parsers are only called if the `@title`, `@description` and `@details` tags are used explicitly, but not if they are just extracted from the first and second paragraphs and the rest.

It would be great if they could be called from `process_description` as well.

gaborcsardi:
To address #364.

### DIFF
--- a/R/parse-preref.R
+++ b/R/parse-preref.R
@@ -78,7 +78,7 @@ process_description <- function(intro, other_pieces) {
     description <- c("description", title[2] %||% piece("title"))
   }
 
-  # Every thing else = details, combined with @details.
+  # Every thing else = details, combined with @details, in parse_elements
   if (length(paragraphs) == 0 || paragraphs == "") paragraphs <- NULL
   if (length(paragraphs) > 0) {
     details <- c("details", paste(paragraphs, collapse = "\n\n"))
--- a/R/parse-preref.R
+++ b/R/parse-preref.R
@@ -15,20 +15,24 @@ parse.preref <- function(lines) {
   ## Compress the escaped delimeters.
   elements <- str_replace_all(elements, fixed("@@"), "@")
 
-  parsed <- parse_elements(elements[-1])
-
-  if (elements[[1]] != "") {
-    check_rd(NULL, elements[[1]])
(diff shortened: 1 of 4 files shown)


gaborcsardi:
Wait please, needs some more testing.

gaborcsardi:
@hadley I think it is good to go now. I run it through a couple of biggish packages, like `devtools` and `igraph`, and the docs are literally the same.

R/parse-preref.R
```diff
@@ -15,20 +15,24 @@ parse.preref <- function(lines) {
...
 
 parse_elements <- function(elements) {
-  pieces <- str_split_fixed(elements, "[[:space:]]+", 2)
+
+  pieces <- str_split_fixed(elements[-1], "[[:space:]]+", 2)
+  desc <- process_description(str_trim(elements[1]), pieces)
+  pieces <- rbind(desc, pieces, deparse.level = 0)
+
+  ## Merge multiple @details tags, one might have come from the intro
+  didx <- which(pieces[,1] == "details")
```
hadley: Isn't that handled by line 82 below?

gaborcsardi: No, the comment is wrong there, actually. It would be slightly more difficult to handle it there. Now `other_pieces` is an input only in `process_description`, but not output. Handling it there would mean that you need to output/change it (remove or update the existing `details` field), and the output value of the function gets more messy. That's why I moved it up here.

gaborcsardi: I'll fix the comment in a sec.

gaborcsardi: Comment fixed.

gaborcsardi:
@hadley could you please take another look? Thanks.

hadley:
Thanks!  Just to double check, this is a refactoring right? Otherwise we'll need a bullet in NEWS.

gaborcsardi:
Yes, actually, this a bug fix in the public API, strictly speaking, so I'll add a bullet to the NEWS.

PR #324, opened 2015-02-23
Parse in collate order

Geoff99:
Hi @hadley 
These changes should ensure that `roxygen2:::parse_package` reads the .R files in collate order when it parses them, and thus resolve issue #323 .
More than happy to make any improvements or corrections needed.

Issue #323, opened 2015-02-22
Proposal : could roxygen2 parse the R files in collate order, to improve @describeIn and @rdname


Geoff99:
#### Synopsis

Could 3 lines of code in roxygen2 be changed so that it **parses** the R files in collate order, when it is generating  documentation?  This would improve the readability of the generated documentation when using `@describeIn` and/or `@rdname` split across several .R files, as may happen when working with S4 classes, generic functions and methods. 

#### Background

The present version of roxygen2, 4.1.0, reads the R code files for a package in alphabetical order when it *parses*  them --- inside `roxygen2:::parse_package()` there is a line of code which says :  

`parsed <- lapply(r_files(base_path), parse_file, env = env)`  

`r_files(base-path)` returns a list of .r filenames in alphabetical order.

In contrast, both `roxygen2::roxygenise()` and `devtools::document()` *source* the files in collate order.

#### Motivation

The consequence is that although the sourced code itself has been loaded in the correct (collate) order, the documentation that roxygen2 generates appears in the alphabetical order of the R files.  This doesn't matter most of the time, because often the contents of a single Rd file come from a single .r file.  

However when `@rdname <some-name>`, or `@describeIn <some-other-name>` have been used in more than one R file, and the *correct* (i.e. collate order) is not alphabetical, it leads to the generated documentation appearing in the *wrong* order.  This typically might happen when using S4 classes, or generic functions and methods.  For example if the method appears in a file which alphabetically precedes the file in which the generic function is created, the generated documentation describes the method first, and the generic function second, *even* if `#' @include ...` has been used to specify the correct collate order for the files.  See the example below.

#### Proposed change

It requires changes to only 3 lines of code inside roxygen2 to have it parse the code files for a package in collate order.  

Two of those lines are in the `roxygen2:::package_files` function (in source.r) and also have the effect of making that function, which is currently only used by `roxygen2:::source_package`, behave more as I suspect is actually intended. At the moment `roxygen2:::package_files`  fails to deduplicate the list of file names that it returns, which means that `roxygen2:::source_package` will source the code files more than once. See the addendum to this example below for an illustration of this.

The third line to be changed is in  `roxygen2:::parse_package` itself.

If you are amenable to considering this change, I can easily submit a pull request. I have written a branched version of roxygen2 already while I was investigating why my documentation didn't look like I expected.

##### File b.r is logically the first file, but alphabetically second.

It creates and documents a generic function, `f_gen`

```{r}
[...]

PR #381, opened 2015-10-01
Parse R files in collation order

krlmlr:
so that the order of documentation that comes from multiple files can be controlled

Looks like package_files() and probably also source_package() was broken before, because r_paths() returns absolute paths. Need to investigate.

@hadley: Do you like the idea?

### DIFF
--- a/R/parse.R
+++ b/R/parse.R
@@ -1,6 +1,6 @@
 parse_package <- function(base_path, load_code) {
   env <- load_code(base_path)
-  parsed <- lapply(r_files(base_path), parse_file, env = env)
+  parsed <- lapply(package_files(base_path), parse_file, env = env)
   blocks <- unlist(parsed, recursive = FALSE)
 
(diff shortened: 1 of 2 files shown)


Geoff99:
Hi there @krlmlr 

I haven't compared the proposed code, but *I* like the idea of solving this - pls see issue #323 and pull request #324.

Cheers

Geoff

krlmlr:
@Geoff99: Sorry, I completely missed that one. I was searching for "collation", and your contributions slipped through.

Geoff99:
No worries @krlmlr 

Just wanted to save some work if possible.  No guarantee the way I "fixed' the parse in collation order issue is the best one.  Happy with any approach that solves it.

Best

Geoff

krlmlr:
Thanks, @Geoff99, you definitely did save work. Your patch is virtually identical to mine, except for the `normalizePath()` call, which I think is a good idea.

PR #392, opened 2015-10-07
Default title and description for package docs

### LINKED ISSUE #349: Proposal: @docType package assigns title and description from the DESCRIPTION file
An `.R` file with the following contents

```
#' @name PackageName
#' @docType package
NULL
```

should create the `PackageName.Rd` with `\title` and `\description` set from the `DESCRIPTION` file. Of course, both can be overridden manually. One can use `#' @details` to add further information.

Use case: Automatic creation of consistent and meaningful package documentation files which are useful, e.g., for data-only packages.

krlmlr:
Using `_PACKAGE` as sentinel value (to avoid unlikely collision with an object actually named "PACKAGE").

~~No functionality yet, just to make sure we're on the same page here.~~

Fixes #349.


R/object.R
```diff
@@ -113,4 +113,6 @@ obj_type.refMethodDef <- function(x) "rcmethod"
 #' @export
 obj_type.function <- function(x) "function"
 #' @export
+obj_type.package_doc <- function(x) "package_doc"
```
hadley: I'd just call it "package" (it's implicitly about documentation)

R/object-from-call.R
```diff
@@ -24,6 +24,10 @@ object_from_call <- function(call, env, block) {
 }
 
 find_data <- function(name, env) {
+  if (name == "_PACKAGE") {
```
hadley: I'd prefer `identical()`, but that's the idea I was thinking of

R/object-from-call.R
```diff
@@ -25,6 +25,14 @@ object_from_call <- function(call, env, block) {
 
 find_data <- function(name, env) {
   ns <- env_namespace(env)
+
+  if (identical(name, "_PACKAGE")) {
+    return(structure(
+      list(pkg_name = packageName(env),
+           base_path = getNamespaceInfo(ns, "path")),
```
krlmlr: Not sure if there's a better way to find the `base_path` here (or in `object_defaults.package()`).

krlmlr:
Works for me.

R/object-defaults.R
```diff
@@ -29,6 +29,20 @@ object_defaults.data <- function(x) {
...
 #' @export
+object_defaults.package <- function(x) {
+  desc <- read.description(file.path(x$value$base_path, "DESCRIPTION"))
+
+  pkg_name <- x$value$pkg_name
+  list(
+    docType = "package",
+    title = as.character(desc$Title),
+    description = as.character(desc$Description),
+    aliases = c("NULL", paste0(pkg_name, "-package")), # See #202
```
hadley: Could you make the comment a bit more direct? i.e. "NULL" prevents addition of default aliases

R/object-from-call.R
hadley: In `object_defaults` you might be able to do some manipulation of `x$srcref$filename`, but I don't know if that contains the full path

krlmlr: In `object_defaults()`, x doesn't have srcref -- only alias, value, and methods. Only the caller, add_defaults(), has access to srcref.

hadley: Oh true, but we could pass that on

krlmlr: Of course: dae13b2. I'm not sure if this is really better, though. Please let me know which you prefer.

R/object-defaults.R
krlmlr: Done.

R/object-from-call.R
hadley: I do prefer it, just because it keeps the division of responsibilities more clear.

krlmlr: Updated, but with modified division of responsibilities: 0a6f850

hadley:
Thanks!

jimhester:
Maybe this is too close to the release to change, I think the underscore prefix for "_PACKAGE" is unnecessarily conservative. How often would someone previously be documenting a _string_ of "PACKAGE". I could see someone documenting a `PACKAGE` object, but never a bare string. Can't we get largely the same collision guarantees by checking the "PACKAGE" object is a `character(1)` in the package environment (At https://github.com/klutometis/roxygen/pull/392/files#diff-00c18366780143a8dfe7c9420de32495R27 perhaps?)

Necessitating the leading underscore just complicates the typical case without a real benefit IMHO.

hadley:
I like the underscore because it makes it clear that this is a special string. We might add others in the future.

jimhester:
Fair enough, consider me overruled :smile:

PR #397, opened 2015-10-11
Allow roxygen comments in object to be documented

krlmlr:
useful e.g. to document implementation details inline

Example:

```
func <- function(a) {
  #' @param a an integer
  stopifnot(is.integer(a))

  #' @details obtw: this is how it works internally...
}
```


hadley:
Hmmmmmmmmm. On one hand, I like the idea for RC/R6 and the implementation is simple. OTOH, I'd prefer to enforce a consistent style of roxygen usage. Let me sit on it for a bit.

krlmlr:
As discussed: Added tests, no off-by-one errors found.

PR #398, opened 2015-10-14
use doxygen if src directory exists  and doxygen is accessible

cbarbu:
Account for the remarks of Hadley on precedent pull request


cbarbu:
I've seen the failure of the Travis CI build, I'll try to fix that.

cbarbu:
That seems fine now. I'm not planning to make any more changes for now.

hadley:
I think this is a fine idea, but I don't like the idea of maintaining a whole bunch of code that I don't really understand. Instead, I'm working towards a system where you'll be able to add put this in your own package.

PR #404, opened 2015-10-20
Fine-tune package doc

krlmlr:
- name = rdname = pkgname-package
- it's necessary to add both pkgname and pkgname-package to the aliases
- the default value for aliases was malformed -- only a single string is
  accepted
- also update vignette to include new _PACKAGE sentinel


hadley:
Thanks!

PR #405, opened 2015-10-21
Fix, whitespace was detected as intro @details

gaborcsardi:
This was added in be2d3ea70075f52de3c6b33471c00161d54c7b87

I forgot a str_trim() and never found it in the tests, because my test docs were tidy. Please merge this before the release.


hadley:
Thanks!

PR #410, opened 2015-10-22
Generate high-level description for data objects

krlmlr:
Currently very simple and very generic. Examples (from tests):

- An object of class \code{list} of length 2.
- An object of class \code{matrix} with 10 rows and 10 columns.
- An object of class \code{data.frame} with 1 rows and 2 columns.
- An object of class \code{array} of dimension 3 x 3 x 3.
- An object of class \code{xtabs} (inherits from \code{table}) of dimension 3 x 2 x 3.

The format-generating function is an exported S3 generic, specialized formatters can be added easily, and packages should be able to supply custom formatters.

In the long run, it might be better to add the format from `block_to_rd()`, and not as default.


hadley:
Thanks!

PR #411, opened 2015-10-22
Allow empty line after @template

krlmlr:
Currently, the trailing `\n` is included in the template name. Is there a better alternative to `trimws()`?


hadley:
roxygen2 depends on stringr, so I use `str_trim()` elsewhere

hadley:
I'm also a bit unsure if this is the right way to solve these problems. It might be better to more carefully define the tag parsers to explicitly either preserve or trim. (You don't need to change this PR, it's more something we should think about in the future)

krlmlr:
Changed. I'll briefly check how much of a problem this is for other tags.

Issue #413, opened 2015-10-22
Improve support for empty lines


krlmlr:
Things start to fall apart in various ways when I randomly insert empty lines in the roc_proc_text() calls in the tests.

As mentioned in https://github.com/klutometis/roxygen/pull/411#issuecomment-150335926.

PR #414, opened 2015-10-22
Demo: Insert empty lines in tests

krlmlr:
Example for #413. Not for merging.


hadley:
I think this would be best fixed by in the parsers. It would also be nice to do a bit more refactoring in their, i.e. changing `.` to `_`, and thinking about whether the names could be more descriptive. Also they could be moved from the bottom of `parse-preref.R` to their own `tag-parsers.R` file.

Issue #418, opened 2015-10-30
Bug in call to tag_warning for noRd


thomasp85:
Parsing @noRd results in a value of "\n". This raises a warning about the tag not supporting parameters. Worse the resulting tag_warning contains a bug:

```r
> tags$noRd
function (x) 
{
    if (x$val != "") {
        tag_warning("has no parameters")  ### Should be tag_warning(x, "has no parameters")
    }
    else {
        x
    }
}
<environment: namespace:roxygen2>
```

Both this and #417 seems to be related to how tags are being parsed. Probably there are a problem with including "\n" for the last line in the tag..? If the inclusion is intentional then the tag parsing should be modified to take this into account.

Most tags include a test for `x$val == ""`. These will never fire as tags will always contain "\n"

PR #419, opened 2015-10-30
Fix wrong call in tag_warning at parse.toggle

thomasp85:
x should be passed as the first parameter. As discussed in #418

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # roxygen2 5.0.0.9000
 
 * `@export` again allows trailing new line (#415).
+* Fixed bug in `@noRd`, where usage would cause error (#418).
 
 # roxygen2 5.0.0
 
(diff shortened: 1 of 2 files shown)


thomasp85:
This only fixes the wrong call. noRd is still throwing warning about parameters and is not getting respected in terms of blocking documentation...

hadley:
Can you fix the other problem too? It just needs `x$val <- str_trim(x$val)`

thomasp85:
Yeah - just needed lunch

thomasp85:
Seems to be fixed now

hadley:
Could you please add a bullet point to NEWs as well?

thomasp85:
Sure - same as the latest version?

hadley:
Yup

thomasp85:
There you go

hadley:
Thanks!

PR #421, opened 2015-10-30
Message was not updated for version change

jimhester:


hadley:
Maybe we could now just drop the version number?

hadley:
It only really mattered when you were upgrading from <3.0.0 to 4.0.0

hadley:
Thanks!

PR #425, opened 2015-11-03
Print source location in error message about missing name

krlmlr:
Otherwise it might be a bit difficult to spot.


hadley:
Thanks!

PR #432, opened 2015-11-09
Modify test for better compatibility with R 3.1.1

krlmlr:
- inline variable to avoid mocking

Fixes hadley/testthat#318.

I'm not sure if this will fix the issue, but could you please supply a session_info()?


hadley:
That seems to do it - thanks!

PR #436, opened 2015-11-23
Never overwrite manually created NAMESPACE

krlmlr:
even when using roxygen2 for the first time.

A follow-up to #348, which works fine except for first-time usage of roxygen2.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # roxygen2 5.0.1.9000
 
+* Never overwrite manually created `NAMESPACE` or documentation files, even if
+  using `roxygen2` for the first time (@krlmlr, #436).
+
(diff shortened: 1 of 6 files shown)


hadley:
Can you give a bit more detail about why this is the correct fix?

krlmlr:
I feel `check` is a misnomer, 3124846 renames the argument so that the action is more explicit.

hadley:
Should I rename `check` to `is_first`? The generic isn't exported (yet) so I think it's fine to change the argument name.

krlmlr:
I'm updating the pull for the current codebase (to use is_first). I'm not entirely sure why the update behavior of roxygen2 should be any different for the first run.

hadley:
Merged by hand in 596893c42b48aaa91781a60ffaad3c2dc0000ac0

PR #437, opened 2015-12-02
Trivial vignette update

dlebauer:
As of 5.0, version is no longer written to .Rd files (thanks!) updated vignette to reflect this.

Perhaps you want to provide more detail about the change as of v 5.0


hadley:
Thanks!

PR #445, opened 2016-01-07
Fix @inheritParams for multi parameter lines

jefferis:
The names of params for multi param lines were not being trimmed after being split.

See 17017b0f8546c19301f1650e3fe03bf789bb72a5 for discussion of alternatives to this particular fix.

### DIFF
--- a/R/inherit-params.R
+++ b/R/inherit-params.R
@@ -71,8 +71,8 @@ find_params <- function(inheritor, topics, name_lookup) {
   param_names <- str_trim(names(params))
   param_names[param_names == "\\dots"] <- "..."
 
-  # Split up compound names on , duplicating their contents
-  individual_names <- strsplit(param_names, ",")
+  # Split up compound names on , (swallowing spaces) duplicating their contents
+  individual_names <- strsplit(param_names, ",[ ]*")
   reps <- vapply(individual_names, length, integer(1))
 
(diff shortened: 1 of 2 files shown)


R/inherit-params.R
```diff
@@ -71,8 +71,8 @@ find_params <- function(inheritor, topics, name_lookup) {
   param_names <- str_trim(names(params))
   param_names[param_names == "\\dots"] <- "..."
 
-  # Split up compound names on , duplicating their contents
-  individual_names <- strsplit(param_names, ",")
+  # Split up compound names on , (swallowing spaces) duplicating their contents
+  individual_names <- strsplit(param_names, ",[ ]*")
```
hadley: Wouldn't `\\s` be more appropriate here?

jefferis: Sure. Or Maybe just remove all whitespace first and then split on `","`?
```r
individual_names <- strsplit(gsub("\\s*","", param_names), ",")
```

hadley: That seems riskier to me. Best to keep the change as small as possible.

jefferis: OK so I've pushed an update with
```r
individual_names <- strsplit(param_names, ",\\s*")
```
thanks for reviewing.

PR #449, opened 2016-01-27
cover simplest case where UseMethod has only one argument in generic,…

### LINKED ISSUE #448: Roxygen doesn't understand UseMethod("generic", x)
Having battled through the R documentation, I don't see any objection to using `UseMethod` within an S3 method to dispatch again, but doing this breaks `roxygen2:::calls_use_method` which simply looks for whether `UseMethod` exists.

I want to do something like:

```
#' my method
#' @export
my_method <- function(x)
  UseMethod(x)

#' @describeIn my_method a class stuff
#' @export
my_method.a_class <- function(x)
 return("something")

#' @describeIn my_method determine class and re-dispatch
#' @export
my_method.default <- function(x)
  class(x) <- "a_class"
  UseMethod("my_method", x)
```
This works from an R point of view, and, although it is not a common use of UseMethod, it is the simplest way I can see to solve a more complicated problem in my package (https://github.com/jackwasey/icd9). Unfortunately, roxygen2 gives:
`Error: Don't know how to describe s3generic in s3generic.`
Commenting the second `UseMethod` lets roxygen run okay.

Incidentally, `NextMethod` is not the solution to this for multiple reasons, not least that you can't update the class passed to NextMethod.

jackwasey:
… but doesn't in re-used UseMethod invocation in S3 methods.

fixes #448

### DIFF
--- a/R/s3.R
+++ b/R/s3.R
@@ -32,8 +32,7 @@ calls_use_method <- function(x) {
   # Base cases
   if (missing(x)) return(FALSE)
   if (!is.call(x)) return(FALSE)
-
-  if (identical(x[[1]], quote(UseMethod))) return(TRUE)
+  if (identical(x[[1]], quote(UseMethod)) && length(x) == 2) return(TRUE)
   if (length(x) == 1) return(FALSE)
   # Recursive case: arguments to call
   for (arg in as.list(x[-1])) {


hadley:
Can you please provide a unit test?

hadley:
This doesn't seem like the right fix to me:

```R
calls_use_method(quote(function(x) {
  UseMethod("x")
}))


calls_use_method(quote(function(x) {
  UseMethod("x", 10)
}))
```

They both already return TRUE

PR #451, opened 2016-01-30
[fix #446] ignore files listed in .Rbuildignore

### LINKED ISSUE #446: Should files listed in .Rbuildignore also be ignored by roxygenise?
It seems that if a `.R` file is listed in the .Rbuildignore, it should be ignored by `roxygenise`.

If you think it's a reasonable idea, I'm willing to submit a PR.

fmichonneau:
Please let me know if you think the approach is reasonable. I can add NEWS entry and attempt tests once this is reviewed. Thanks.


hadley:
@jimhester any chance #458 also covers this?

R/utils.R
```diff
@@ -123,10 +123,39 @@ same_contents <- function(path, contents) {
...
 
-r_files <- function(path) {
+all_r_files <- function(path) {
   sort_c(dir(file.path(path, "R"), "[.Rr]$", full.names = TRUE))
 }
 
+r_files <- function(path) {
+    rbuildignore <- file.path(path, ".Rbuildignore")
+    rfiles <- all_r_files(path)
+    if (file.exists(rbuildignore)) {
```
hadley: This should to be pulled out into a separate function, called something like `ignore_files`

hadley:
Are you still interested in this PR?

jimhester:
It doesn't look like #458 `tools::list_files_with_type` checks `.Rbuildignore`, so won't help this unfortunately.

fmichonneau:
yes, I'll update it before the end of the week if that works for you.

fmichonneau:
I pulled out the code that ignore files into its own function as requested. I also wrote a test and added it to the NEWS file.

R/utils.R
```diff
@@ -96,10 +96,47 @@ same_contents <- function(path, contents) {
...
 }
 
-r_files <- function(path) {
+all_r_files <- function(path) {
   sort_c(dir(file.path(path, "R"), "[.Rr]$", full.names = TRUE))
 }
 
+ignore_files <- function(path, rfiles) {
+    rbuildignore <- file.path(path, ".Rbuildignore")
+     if (file.exists(rbuildignore)) {
```
hadley: It would make the code simpler to flip the check and return early if there is no `.Rbuildignore`

R/utils.R
```diff
@@ -96,10 +96,47 @@ same_contents <- function(path, contents) {
...
+        relative_rfiles <- sub(normalizePath(path), "", normalizePath(rfiles))
+        ## If user adds trailing / when specifying path, we need to remove it.
+        relative_rfiles <- sub(paste0("^(", .Platform$file.sep, "|/)*"), "",
+                               relative_rfiles)
+        ## Figure out if any of the regular expressions matches files
+        ## in R/ Use perl=TRUE as R-exts manual says that patterns
+        ## should be perl-like regular expressions
+        idx_rfiles_to_ignore <- unique(unlist(lapply(rbuildignore_patterns,
+                                                     function(x) {
+             grep(x, relative_rfiles, perl = TRUE)
```
hadley: Would be better to use `grepl` here

fmichonneau:
Thanks for the feedback Hadley, I updated the code accordingly.

In the process I discovered that this approach doesn't work if collation is used by the package. Indeed, the files listed in the Collate field are appended to the files returned by `r_files()` in `package_files()`. Therefore I moved where the ignored files are removed to `package_files()`.

Another approach would be to check for the content of the .Rbuildignore file when the Collate field is generated by updating the code for `generate_collate()`.

R/utils.R
```diff
@@ -100,6 +100,33 @@ r_files <- function(path) {
   sort_c(dir(file.path(path, "R"), "[.Rr]$", full.names = TRUE))
 }
 
+ignore_files <- function(path, rfiles) {
+    rbuildignore <- file.path(path, ".Rbuildignore")
+    if (!file.exists(rbuildignore)) {
+        rfiles
```
hadley: It would be cleaner to `return(rfiles)` here so that the big block below doesn't need to be indented

R/utils.R
```diff
@@ -100,6 +100,33 @@ r_files <- function(path) {
...
+        ## Figure out if any of the regular expressions matches files
+        ## in R/ Use perl=TRUE as R-exts manual says that patterns
+        ## should be perl-like regular expressions
+        idx_rfiles_to_ignore <- Reduce("|", lapply(rbuildignore_patterns,
+                                       grepl, relative_rfiles, perl = TRUE))
+        rfiles[!idx_rfiles_to_ignore]
+    }
+}
+
+dots <- function(...) {
```
hadley: I think this was included by accident

R/utils.R
```diff
@@ -100,6 +100,33 @@ r_files <- function(path) {
...
+        ## leading up to the path. `normalizePath` ensures that the
+        ## path is properly expanded.
+        relative_rfiles <- sub(normalizePath(path), "", normalizePath(rfiles))
+        ## If user adds trailing / when specifying path, we need to remove it.
+        relative_rfiles <- sub(paste0("^(", .Platform$file.sep, "|/)*"), "",
+                               relative_rfiles)
+        ## Figure out if any of the regular expressions matches files
+        ## in R/ Use perl=TRUE as R-exts manual says that patterns
+        ## should be perl-like regular expressions
+        idx_rfiles_to_ignore <- Reduce("|", lapply(rbuildignore_patterns,
```
hadley: I think this would be a little easier to read if spread across two expressions

hadley:
Looking good - just a few minor changes left

R/utils.R
```diff
@@ -100,6 +100,33 @@ r_files <- function(path) {
   sort_c(dir(file.path(path, "R"), "[.Rr]$", full.names = TRUE))
 }
 
+ignore_files <- function(path, rfiles) {
```
hadley: I think the arguments should be swapped: the primary thing being modified is the vector of paths to R files

PR #458, opened 2016-02-29
Roxygen handles code in OS specific subdirectories

jimhester:
This PR basically just uses `tools::list_files_with_type` which handles files in os specific directories. This gets us at least part of the way to handling OS specific subdirectories.

Things still missing would be support for separate `Collate.unix` and `Collate.windows` directives and writing documentation to `man/unix/` and `man/windows`. These features aren't widely used and I am also unsure what the roclet syntax should look like, so I did not try to tackle that.

### DIFF
--- a/R/roclet-collate.R
+++ b/R/roclet-collate.R
@@ -53,7 +53,7 @@ update_collate <- function(base_path) {
 }
 
 generate_collate <- function(base_path) {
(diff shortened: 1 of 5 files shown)


hadley:
Could you add a bullet to NEWS please?

hadley:
Let's not worry about this for now

Issue #388, opened 2015-10-06
Documentation for R6 methods


hadley:
Should look something like this:

---
# Methods
## Method 1

```
class$method(a, b, c = 12)
```

Description.
### Parameters
- Param 1
- Param 2
- Param 3
### Returns

What the method returns
### Example

``` R
a <- 10
b <- 5
a + b
#> [1]
```

---

Rd would look something like:

``` latex
\section{Methods}{
\subsection{\code{method1}}{

\code{class$method1(a, b, c = 12}

\subsection{Parameter}{

}
}
```

and similarly for fields. 

Parameters, returns, and methods would be optional. If no methods use them, just list usage + (optional) description with bullets.  Can also use top-level `@examples` if you want to put all examples in one place.

hadley:
It's going to be a lot of work to fix this, and it's an incredibly special case, so I'm closing because realistically this will never get to the top of my todo list.

krlmlr:
How about also parsing roxygen comments *inside* the objects that are documented? They seem to be available in the [`parsed` list](https://github.com/klutometis/roxygen/blob/1b8ea387dacc22d023a108a219f6614e76469df6/R/parse.R#L24). Would be useful also for DBItest.

krlmlr:
Why not one `.Rd` file per method? The list of methods could be autogenerated, with links to detailed methods documentation if available.

```
\section{Public methods}{
  \itemize{
    \item{\code{initialize(a, b, c = 2)}
    \item{\code{\link[=Class$method1]{method1}(d, e)}: Title of method documentation
  }
}
```

hadley:
I don't think you want want file per method, because with this style of OO you tend to have lots and lots of little methods. Most docs for js, ruby, python, and java document all methods on a single page.

Enchufa2:
Hi! If I may make a couple of suggestions, it would be great something like this (I don't know if I'm defining too many subsections):

# Class
## Description
blablabla
## Inherits
BaseClassLink
## Usage
Class$new(param1, ...)
## Arguments
* param1
* ...

## Public attributes
* attr1
* ...

## Public methods
your structure here
## Active bindings
### ActiveBinding1
```R
class$activebinding1
class$activebinding1 <- value
```
Description.
#### Parameters
* value

etc. (same structure as methods)

## Details
details

## References
refs

## See also
blablabla

## Examples
General examples

levithatcher:
[...]

PR #465, opened 2016-04-08
Added functionality to document R6classes using docblocks that allow …

hhoeflin:
Wanted to make a suggestion on how to document R6 classes.

It is mostly in its basic functionality copied from how roxygen handles rc classes, but expanded from there on how the documentation works.

In each R6class-method, it still uses docstrings for documentation, like RC classes. However, instead of a single string, it also accepts several of them - i.e. a docblock as several docstrings.

Inside the docblock, @param lines are allowed that specify parameters.

In the documentation, the methods are set as follows:

<method name>  *Usage:*
.... usage string of method ...

Description of method

If parameters were specified, this is followed by a 
*Parameters:*
\describe-block with one \item per parameter. 

It keeps the documentation for all methods together in a block and allows for more detail on how to document the parameters. usually, the documentation of each method will still be only 5 lines or so, so while it gets longer, it isn't excessively so.


hhoeflin:
Documentation would look something like this

[trialr6.txt](https://github.com/klutometis/roxygen/files/210408/trialr6.txt)

and the resulting Rd file like this

[Myclass-class.txt](https://github.com/klutometis/roxygen/files/210409/Myclass-class.txt)

dselivanov:
Why this can't be merged?

epicfarmer:
This commit doesn't allow for \code or other similar commands inside the docstring.  Is there some reason its impossible to continue using the #' to denote roxygen comments?

hadley:
Please discuss in #388

PR #493, opened 2016-08-08
Tidy filename Rd display from roxy_field_backref

LiNk-NY:
When multiple files are referenced in one documentation page, they will run off the page. 

This will create a one line per file format for the .Rd display. Example below:
```
cat(sprintf("%% Please edit in:\n%% %s\n",
      paste(c("a", "b", "c"), collapse = ",\n% ")))
```


hadley:
I'd rather just wrap to (say) 80 characters

LiNk-NY:
That works too

hadley:
Do you want to have a go at implementing that?

LiNk-NY:
Yes, should have it soon..

LiNk-NY:
I didn't find an easy way to add "\n" to the output of `strwrap`. I ended up using paste twice. https://github.com/LiNk-NY/roxygen/blob/master/R/rd-tag-api.R#L97

LiNk-NY:
Looks ready to go unless you have a more elegant way of adding "\n" to the end of a stri_wrap return

hadley:
Thanks!

Issue #505, opened 2016-09-14
Better links in markdown


gaborcsardi:
From this discussion: https://github.com/klutometis/roxygen/pull/499/files/fc8ab2eddd1f855f7841a94151aaaeefdeae4b10#r78301946

Plan: 
* link to function in the same package: `[fun]`
* link to function in another package: `[pkg::fun]`
* link to function with link text: `[text][fun]`
* link to function in another package with link text: `[text][pkg::fun]`

I think this is all possible by simply adding the link references to the end of the text.

gaborcsardi:
I am proposing that I would do this, obviously.

hadley:
I like it!

gaborcsardi:
Closed by #506.

PR #506, opened 2016-09-16
Easier links in Markdown

gaborcsardi:
### EDIT: see current syntax below, here: https://github.com/klutometis/roxygen/pull/506#issuecomment-247879044

~~The old syntax is not supported. The current syntax:~~
 * ~~[function] becomes \code{\link{function}}~~
 * ~~[pkg::function] becomes \code{\link[pkg]{function}}~~
 * ~~[name][dest] becomes \link[=dest]{name}~~
 * ~~[name][pkg::bar] becomes \link[pkg:bar]{name}~~


gaborcsardi:
For #505.

I also updated the roxygen2 docs, and there are no changes in the Rd files (except for one mistake that I fixed).

You can do everything with the new syntax, so it does not make sense to keep the old syntax.

hadley:
And now you can do `` `[fun]()` `` right?

gaborcsardi:
Good question, I did not think about that. So, why do you want to add the `()`, to denote that it is a function, or just as part of the markdown link?

hadley:
I like to do that to denote that it's a function rather than a variable or argument.

gaborcsardi:
Got it. Unfortunately `` `[func]()` `` does not work. I could make it work, but it'll be awkward, the `()` won't be part of the link.

How about just `[func()]`? This does not work currently, either, but I think it is
* easier to write,
* easier for me to implement, and 
* the `()` is part of the link.

gaborcsardi:
Btw. `[func()]` is `\code` by default, no need for backticks.

hadley:
Just to confirm:

* `[func]` -> `\link{func}`
* `[func()]` -> \code{\link[func]{func()}}`

?

The reason I have put the parens outside the link in the past is that it's easier to write `\code{\link{func}()}`

gaborcsardi:
We already have `[func]` -> `\code{\link{func}}`.

I am proposing to add `[func()]` -> `\code{\link[=func]{func()}}`

Hopefully the last one is OK with Rd, and it will be shown as code. If not, I'll do it your way: `\code{\link{func}()}`

But nevertheless the markdown will be `[func()]` I think.

hadley:
Hmmmmm. I'm not sure about `[fun]` automatically adding code. Admittedly that is what you do 95% of the time, but occasionally you do want to link to a general topic (e.g. `\link{Quotes}`) that shouldn't be code formatted.

gaborcsardi:
How about having another notation for that?

Btw you can already do [Quotes][Quotes] which is ugly but we can have some
shortcut for it. Eg [][Quotes] or [Quotes][]. Not very intuitieve though.

gaborcsardi:
Another idea. How about automatically adding code for `[fun()]`, but not for `[object]` ?

hadley:
Yeah, that's what I was thinking

gaborcsardi:
Cool, I'll update soon(ish), maybe today.

gaborcsardi:
Updated. The current syntax:
 * `[fun()]`            becomes `\code{\link[=fun]{fun()}}`
 * `[obj]`              becomes `\link{obj}`
 * `[pkg::fun()]`       becomes `\code{\link[pkg:fun]{pkg::fun()}}`
 * `[pkg::obj]`         becomes `\link[pkg:obj]{pkg::obj}`
 * `[text][fun()]`      becomes `\link[=fun]{text}`
 * `[text][obj]`        becomes `\link[=obj]{text}`
 * `[text][pkg::fun()]` becomes `\link[pkg:fun]{text}`
 * `[text][pkg::obj]`   becomes `\link[pkg:obj]{text}`

It was not so trivial, but this should work fine I think.

hadley:
Thanks!

Issue #507, opened 2016-09-16
Global roxygen options for a package


gaborcsardi:
We would put them in the `DESCRIPTION` file, in the `Roxygen` field, which is allowed, but not used currently. 

The plan is to have R code there, and initially we would support switching Markdown parsing on/off:
```
Roxygen: list(markdown = TRUE)
```

I will submit a PR in the coming days.

gaborcsardi:
I see you are finishing things up. I'll try to write this tomorrow.

hadley:
I'd also consider turning markdown on by default, and we can see how it goes when applied to my packages.

gaborcsardi:
Of course you would turn it on, because you can blame me if sg goes wrong.
:)

With the current link implementation, there will be problems with Rd
brackets for example.

Why don't we just turn it on for your packages?

Regardless, we need this global switch, I think. I'll write it tonight.

gaborcsardi:
Wait a sec, there is code for this already: https://github.com/klutometis/roxygen/blob/596893c42b48aaa91781a60ffaad3c2dc0000ac0/R/roxygenize.R#L86

I am sorry, I wasn't aware of this. So I can just all markdown here.....

gaborcsardi:
This can be closed, there is a way, the `Roxygen` field. It works fine for markdown.

PR #515, opened 2016-09-22
Global option for markdown

gaborcsardi:
For #507 

We read the options via the usual `load_options()` in `roxygenize()`, and then push them down the call chain (almost) everywhere. `parse_text()` also gets a `global_options` argument.

In each function where I added the `global_options` argument, I also added a default value (`list()`), so that we don't break other packages.

While this is done, and works, I don't think it is ideal. The best would be to attach the options to some `package` object, and make this reachable from `block` objects. Roxygen does not really have the organization to implement this cleanly, AFAICT.

Another option would be a global options store, but my problem with that is that it is hard to invalidate, and that these options really belong to the package and the blocks, not to roxygen itself.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -94,7 +94,7 @@
 
 * Most fields can now be written using Markdown markup instead of the
   traditional Rd language. See the 'markdown' vignette for details
-  (#364, #431), by @gaborcsardi
+  (#364, #431, #499, #506, #507), by @gaborcsardi
 
 # roxygen2 5.0.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -92,8 +92,6 @@
 
 * Exported `roxy_tag()`, `roxy_tag_warning()` and 
 
-# roxygen2 5.0.1
-
 * Most fields can now be written using Markdown markup instead of the
   traditional Rd language. See the 'markdown' vignette for details
   (#364, #431), by @gaborcsardi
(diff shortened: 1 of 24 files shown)


NEWS.md
```diff
@@ -92,11 +92,9 @@
 
 * Exported `roxy_tag()`, `roxy_tag_warning()` and 
 
-# roxygen2 5.0.1
-
 * Most fields can now be written using Markdown markup instead of the
   traditional Rd language. See the 'markdown' vignette for details
-  (#364, #431), by @gaborcsardi
+  (#364, #431, #499, #506, #507), by @gaborcsardi
```
hadley: Can you mention the global flag and `@md`/`@noMd` here too please?

hadley:
I agree it's not the ideal approach but it works and doesn't need extensive changes to roxygen2.  If you make that one change to NEWS.md, I'll merge

gaborcsardi:
Sure. You want another bullet point or in the same point?

NEWS.md
gaborcsardi: OK, so here. :) GH email notifications lose the context sometimes.

gaborcsardi:
Btw. I'll also update `desc` tonight to get rid of the annoying `Collate` formatting and the trailing space craziness.

hadley:
Thanks!

PR #529, opened 2016-10-21
tag_two_part first part is not parsed as markdown

### LINKED ISSUE #519: Markdown: backticks in first arg of \arguments \item should not be picked up?
https://github.com/rstats-db/DBI/commit/0bf8ce714e2d2c83b71e2f980b3d0dad9d6a06cc

gaborcsardi:
Because that is typically just a name, and it parsing it
causes problems in some rare cases, see #519.
This commit closes #519.

### DIFF
--- a/R/tag.R
+++ b/R/tag.R
@@ -122,9 +122,16 @@ tag_name <- function(x) {
 #'   (FALSE)?
 tag_two_part <- function(first, second, required = TRUE) {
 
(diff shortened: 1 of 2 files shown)


gaborcsardi:
@hadley I am not entirely sure about the purpose of the `tag_*` functions, so please take a look if this change is appropriate.

PR #531, opened 2016-10-23
Export objects via name vector

egnha:
Tag `@exportNames` allows one to export objects using a vector of names. Example:

``` r
fs <- lapply(1:10, function(a) {function(x) a + x})
names(fs) <- paste0("f_", 1:10)
for (nm in names(fs)) assign(nm, fs[[nm]])

#' @exportNames names(fs)
NULL
```

Such a tag is handy if one needs to export batches of closures, say (e.g., in a DSL).

### DIFF
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -105,7 +105,10 @@ ns_S3method          <- function(tag, block) {
 }
 ns_exportClass       <- function(tag, block) export_class(tag)
 ns_exportMethod      <- function(tag, block) export_s4_method(tag)
-ns_exportNames       <- function(tag, block) fun_args("export", tag)
+ns_exportNames       <- function(tag, block) {
+  nms <- eval(parse(text = tag))
+  fun_args("export", nms)
+}
 ns_exportPattern     <- function(tag, block) one_per_line("exportPattern", tag)
 ns_import            <- function(tag, block) one_per_line("import", tag)
 ns_importFrom        <- function(tag, block) repeat_first("importFrom", tag)
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -1,6 +1,6 @@
-ns_tags <- c('export', 'exportClass', 'exportMethod', 'exportPattern',
-  'rawNamespace', 'S3method', 'import', 'importFrom', 'importClassesFrom',
-  'importMethodsFrom', 'useDynLib')
+ns_tags <- c('export', 'exportClass', 'exportMethod', 'exportNames',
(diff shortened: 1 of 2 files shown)


hadley:
Can you give me more of an example where you'd use this?

egnha:
My experience with writing R packages is admittedly rather limitied, but here's one specific example from ML that I think is not too contrived. One might want to provide probability estimators for a bunch of binary classifiers in a programmatic manner, according to a common pattern. It might then make sense to write a single unified doc for the family of estimators, and then just one doc line to do the export. Then something like `@exportNames` could eliminate tedium and boilerplate code.

To be more concrete:

```r
# Suppose `mdl_params` is a list of model-specific parameters, named by S3 method
# Suppose also that `make_estimator()` makes estimators out of those parameters
estprob <- function(x, ...) UseMethod("estprob")
for (mdl in names(mdl_params))
  assign(paste0("estprob.", mdl), make_estimator(mdl_params[[mdl]]))

#' @exportNames paste0("estprob.", names(mdl_params))
NULL
```

Technically, that won't work in the current implementation of `@exportNames` (as I now realize!), since one would then want to generate `S3method()` lines in `NAMESPACE`, rather than `export()`'s. But `@exportNames` could be modified to handle that. I think the basic use case is valid.

hadley:
I'd rather provide `@evalNamespace` by analog to `@evalRd`. That's a bit more general and would still solve your problem.

egnha:
Interesting suggestion. But I'm not sure I understand what you have in mind, exactly. Here's my guess, please correct me if I'm wrong: `@evalNamespace` would allow you to reproduce

```r
#' @exportNames names(fs)
NULL
```

with

```r
#' @export names(fs)
#' @evalNamespace
NULL
```

Another possibility would be to allow `@evalNamespace` to selectively evaluate part of a tag, while inserting the rest of the tag literally, perhaps by quoting the part to evaluate. Something like

```r
f_nms <- paste0(names(fs), collapse = ",")
#' @evalNamespace export(~f_nms)
NULL
```

But having to put the commas in by hand feels clumsy.

hadley:
`@evalNamespace code(...)` would just evaluate `code` and insert the results into the `NAMESPACE`

egnha:
If you still think `@evalNamespace code(...)` would be a useful addition to roxygen, and if I can figure out a reasonable way to do what you've suggested, would it be better in the future for me to post an issue or a (separate) pull request?

hadley:
A new PR or issue would be best

PR #535, opened 2016-11-16
Inherit title

krlmlr:
This enables inheriting the title of a documentation entry, so that an `@inherit foo` is sufficient to document an object. I'm thinking about using it for RSQLite (to inherit method description from DBI), not sure yet if this is the right approach.


hadley:
This seems reasonable to me. Can you please mention in the news bullet?

krlmlr:
Done, also fixed failing test.

krlmlr:
@hadley: PTAL.

hadley:
Thanks!

PR #539, opened 2016-11-30
Fix inheritance of subsections.

### LINKED ISSUE #521: Section inheritance produces slightly wrong formatting
Example: Comparison of original formatting and formatting produced by inheritance. https://github.com/rstats-db/DBI/commit/a9277a4f5a4187ef5f515c9288d9ad12c7ed7a52#commitcomment-19266464

krlmlr:
I believe this is an upstream bug in `tools:::as.character.Rd()`.
The fix here uses a copy of this function under a different name.
Filed [an issue](https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17188) with the R bug tracker.

Fixes #521.

### DIFF
--- a/R/utils-rd.R
+++ b/R/utils-rd.R
@@ -160,17 +160,9 @@ as_character_Rd <- function(x) {
       result <- c()
     }
     else {
-      if (deparse) {
-        dep <- deparseRdElement(as.character(x), c(state,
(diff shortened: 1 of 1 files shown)


krlmlr:
Issue has been just fixed upstream, hopefully will appear in R 3.3.3? For now, I can use this hack or R-devel.

Issue #548, opened 2016-12-24
Provide suitable default filename for objects with names starting with a dot


renozao:
The default filename for objects whose name start with a dot is invalid and skipped/deleted.
Is this on purpose?
Adding the following step to `nice_name` would solve the issue of having to provide a `@rdname` for those objects:
```R
x <- str_replace_all(x, "^\\.", "dot-")
```

There always is `@keywords internal` for the cases where one wants to document an object but does not want the man page to be listed.

hadley:
Generally, by convention, it's bad form to have an exported object that starts with a dot. But I'd consider a PR.

renozao:
The use case is not for exported objects, but for documenting internal objects.

PR #549, opened 2016-12-28
Provides valid default name for object named with a starting dot

renozao:
Minor change to solve issue #548


hadley:
Can you please also include a bullet in NEWS.md?

hadley:
Sorry I missed this for the last release. Would you mind moving the news bullet up to the top?

PR #552, opened 2017-01-03
MD: comment that links to operators fail, closes #550

### LINKED ISSUE #550: How to link to operators defined in other packages using Markdown?
e.g., `magrittr::%>%`, as in `\code{\link[magrittr]{\%>\%}}`?

CC @gaborcsardi.

gaborcsardi:
I also fixed the markdown.html file, that I messed up last time. 

(Btw. is there an easy way to build all files from the `Rmd` vignette?)


hadley:
There's a vignettes roclet that should do it, but I'm not sure if the semantics are exactly right. At some point I need to sit down and think through exactly how and when vignettes should get built.

PR #557, opened 2017-01-06
Filter by relative path but return aboslute paths

jeroenooms:
Fixes https://github.com/klutometis/roxygen/issues/556


hadley:
Thanks!

PR #561, opened 2017-01-16
Interpret `@evalRd` results as Markdown

krlmlr:
Do we want this? Happy to add a test.

### DIFF
--- a/R/rd.R
+++ b/R/rd.R
@@ -359,7 +359,7 @@ topic_add_eval_rd <- function(topic, block, env) {
     tryCatch({
       expr <- parse(text = tag)
       out <- eval(expr, envir = env)
-      topic$add_simple_field("rawRd", as.character(out))
+      topic$add_simple_field("rawRd", full_markdown(as.character(out)))
     }, error = function(e) {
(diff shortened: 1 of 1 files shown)


hadley:
Hmmmm, feels like that maybe should be a separate tag?

krlmlr:
I considered that, but Markdown is currently opt-in anyway, so we should be on the safe side. I agree that the tag's name is misleading, though. Is it worth to add extra complexity here?

gaborcsardi:
I don't know is this is needed, but I think that the patch is not correct, because it does not consider `@noMd` or the global `markdown` option.

krlmlr:
I thought `full_markdown()` -> `markdown()` would look for that option by checking `markdown_on()` ?

hadley:
I think the problem with making it unconditional is that the writer of the function (i.e. `@evalRd foo::bar()`) might be different to the package author (and hence doesn't know whether or not markdown will be enabled).

gaborcsardi:
@krlmlr Yes, you are right, actually.

krlmlr:
Closing for now, the user can use `@evalRd roxygen2:::full_markdown(...)` .

Issue #583, opened 2017-02-08
Add bibliography support


renozao:
This is is to simplify the usage of tag `@references`, using Bibtex files and tags `@cite`, `@bibliography` and command `\cite` within other tag descriptions.

Submitting a pull request towards this now.

hadley:
See PR in #584

hadley:
Closing this because I now believe this is big enough that it belongs in its own package.

PR #584, opened 2017-02-08
Add bibliography support (issue klutometis/roxygen#583)

renozao:
This is a description of the new feature, as detailed in the `rd.Rmd` vignette:

## Bibliography support

References to published material can be documented in man pages, by
manually copy the relevant citations in tags `@references`.
However, citing articles can be automated using Bibtex files, tags

`@bibliography` and `@cite`, and command `\cite`:
 
  * Use tag `@bibliography` anywhere in your code to point roxygen to a local
    Bibtex file, which is typically your personal Bibtex library that contains
    citations for all the material you ever read.
    You can have multiple `@bibliography` tags.
  * A `@cite bibkey` in a block is equivalent to having a tag
    `@references full citation of Bibtex entry for key bibkey`.
  * Commands `\cite{bibkey}` found in any block introductions, tags
    `@description`, `@details`, `@section` (content) or `@param` (description),
    are substituted by strings like `Doe et al. (2000)` and include associated 
    `@references` tags with the full citation.

By default, parsed Bibtex entries will be looked in file `inst/REFERENCES.bib`,
and if not found in the files declared with tags `@bibliography`. 
At the end, the file `REFERENCES.bib` is updated with those new entries.


hadley:
I think this is basically good idea but unfortunately this PR is a long way away from mergeable code. If you are still interested in working on this, I think the place to start would be with a simple implementation of `@cite` relying on a bibliography found in a fixed and known location (i.e. don't support `\cite` and don't support `@references)`. That would majorly simplify the implementation.

Additionally, R6 is not necessary here. Since we don't need a mutable object it would be better to use S3.

Issue #601, opened 2017-03-02
Namespace directive order should honour Collate order


renozao:
Currently, the namespace roclet sorts the directives, which causes check and loading warnings of the following sort:

```
Found the following significant warnings:
  Warning: replacing previous import ‘Rcpp::prompt’ by ‘utils::prompt’ when loading ‘pkgA’
```

These warnings can be avoided by forcing package import order in the `NAMESPACE` file.
However, this is not possible when this file is generated by roxygen, because of the default sort. 
Users should be able to control order by having the `@import` tags parsed in the correct order, which can be achieved easily if the namespace directives are output directly without sorting.

That is having :
```R
roclet_process.roclet_namespace <- function(x, parsed, base_path,
                                            global_options = list()) {
  ns <- unlist(lapply(parsed$blocks, block_to_ns)) %||% character()
  unique(ns)
}
``` 
instead of
```R
roclet_process.roclet_namespace <- function(x, parsed, base_path,
                                            global_options = list()) {
  ns <- unlist(lapply(parsed$blocks, block_to_ns)) %||% character()
  sort_c(unique(ns))
}
```

gaborcsardi:
What if someone wants a NAMESPACE import order that is different from the Collate order? Or that doesn't happen?

Btw. I think a good workaround is to **not** import conflicting names, but refer to them with `utils::prompt` and `Rcpp::prompt`. This is also more readable, as the reader does not have to think where a given reference to `prompt` is coming from.

renozao:
Currently there is no choice in the order. Following Collate order does not really impose any specific order, because the user decides in which file to place the `@import` tags and can choose the Collate order. The point is that it enables the user to fully control the order: any order can be achieved.

It looks like that most of these are not real conflicts, but S4 generics defined for existing S3 generics which I think are supposed to be fine.
I feel that the workaround of specific import is a bit too much for base packages such as `utils` and `stats`, from which you often need many many functions.
I agree with the explicit namespace calling if there really is a conflict.

While sorting directive makes the NAMESPACE look more tidy, I think this should be a secondary objective, given that the order does matter.

gaborcsardi:
> While sorting directive makes the NAMESPACE look more tidy, I think this should be a secondary objective, given that the order does matter.

That's a good point. Maybe this could be opt-in, because otherwise roxygen will re-order the `NAMESPACE` for everybody, which is annoying.

renozao:
I'd even say why allowing for sorting at all?

[...]

PR #604, opened 2017-03-07
Feature/unsorted namespace

renozao:
Proposed solution for #601.
Add global option "namespace_unsorted", to enable keeping Collate order in NAMESPACE statement.


hadley:
This just doesn't seem worth it to me, even as an option.

Issue #628, opened 2017-05-09
Line break in Markdown link text deletes whitespace between words


jennybc:
If line wrapping causes a line break in the text of a Markdown link, the associated space between words goes missing. This does not happen with Rd syntax.

Example of the roxygen comment:

```r
#' Generate a request for the Sheets v4 API
#'
#' Generate a request, using some knowledge of the [Sheets v4
#' API](https://developers.google.com/sheets/api/reference/rest/). Blah blah
#' \href{https://developers.google.com/sheets/api/reference/rest/}{Sheets v4
#' API} Most users should, instead, use higher-level wrappers that facilitate
#' common tasks, such as reading the contents of a sheet. The functions here are
#' intended for internal use and for programming around the Sheets API.
```

Screenshot of the help. Note "Sheets v4API" (no space, from Markdown) versus "Sheets v4 API" (with space, from Rd), despite a line break falling in exactly the same place.

<img width="633" alt="screen shot 2017-05-09 at 3 01 13 pm" src="https://cloud.githubusercontent.com/assets/599454/25874672/6c72d68c-34c8-11e7-9100-4c9a7ad6672e.png">

egnha:
Here's a somewhat shorter demonstration of the problem:
```R
out <- roc_proc_text(rd_roclet(), "
  #' Title
  #'
  #' Link text [broken
  #' across lines](url) doesn't preserve whitespace.
  #' @md
  foo <- function() {}")[[1]]
```
gives
```
% Generated by roxygen2: do not edit by hand
% Please edit documentation in Rtmpp2fxtt/fileb6077549a9
\name{foo}
\alias{foo}
\title{Title}
\usage{
foo()
}
\description{
Link text \href{url}{brokenacross lines} doesn't preserve
whitespace.
}
```

egnha:
Unwinding the (internal) function `markdown()`, it looks like the culprit is the call `xml_text(xml)` in the `link` method in [`markdown_tags`](https://github.com/klutometis/roxygen/blob/master/R/markdown.R#L185):
```R
invisible(markdown_on(TRUE))
markdown("[broken\nacross lines](URL)", markdown_tags)
#> [1] "\\href{URL}{brokenacross lines}"
```
It seems that the newline isn't preserved as whitespace because when `markdown()` produces the final link text by calling
```R
xml_text(xml)
```
on the XML object
```R
```
"
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #630, opened 2017-05-17
Preserve whitespace between words in markdown link text (#628)

egnha:
Implements the patch proposed in issue #628.


R/markdown.R
```diff
@@ -182,7 +182,7 @@ markdown_tags <- list(
     } else if (dest == "" || dest == xml_text(xml)) {
       list("\\url{", xml_text(xml), "}")
     } else {
-      list("\\href{", dest, "}{", xml_text(xml), "}")
+      list("\\href{", dest, "}{", xml_link_text(xml), "}")
```
gaborcsardi: Can you please call it `xml_href_text` ?

gaborcsardi:
Thanks much, it looks great!

R/markdown.R
egnha: Sure thing.

egnha:
My pleasure. :)

gaborcsardi:
Actually, I wonder if we should do this everywhere. Converting a line break to a space seems like a good default behavior. 

(You don't need to modify this PR, just thinking aloud.)

hadley:
Could you please add a bullet to NEWS?

hadley:
Thanks!

For future reference, please make sure that one of the commit messages includes `Fixes #628` so the accompanying issues gets automatically closed when the PR is merged.

egnha:
I'll keep that point in mind for next time. Thanks for the review and merge!

Issue #273, opened 2014-08-13
Add support for single files


klmr:
At the moment roxygen2 seems to only work for packages. However, not all code comes in the form of packages. Even though there’s an overwhelming consensus to use packages even for small analysis projects, it seems unnecessary to push this preference onto the user, since roxygen2 contains all the necessary functionality already, it just needs to expose it in its API.

There are at least two concrete use-cases for this (the second being relevant for me):
- [Single-file one-off Markdown reports](http://stackoverflow.com/q/18923405/1968) generated with [knitr](/yihui/knitr) or Sweave. Even though such reports can generally live without documentation, having it would still help during writing, and would promote writing better code from the start, which could later be put into a package
- [modules](/klmr/modules), a modern alternative to conventional R packages, which organises reusable code units in a fashion similar to Python and [JavaScript](http://nodejs.org/api/modules.html). The feature most sorely lacking from modules at the moment is support for documentation, and it would not make sense to reinvent the wheel, when there’s a perfectly fine tool out there.

klmr:
… and now Roxygen2 breaks ‹modules› because `roxygen:::parse_file` was removed. 👎

misterhenson:
In fact this is highly useful, if you have to do technical documentation on a project, just saying.

klmr:
@misterhenson If you’re interested in this, see #634, especially the links to code in the comments there, where I’m implementing this.

PR #634, opened 2017-05-29
Support non-package code

klmr:
As suggested by @hadley in #273, here’s a suggested API for trivially adding single-file parsing support for Roxygen2.

At the moment the proposed change is pretty minimal (but already sufficient for my purposes): it adds a single, exported function `parse_code` that takes a file or a piece of R code as text, along with an environment of parsed R expressions, and parses the Roxygen tags associated with these objects, to be returned to the user.

I could alternatively/additionally imagine providing a `roxigenize_file` function (as shown in branch [`single-file-support-roxygenize`](../compare/master...klmr:single-file-support-roxygenize?expand=1)).

I would greatly appreciate feedback to make this PR suitable for merging into Roxygen2.

Here’s why I think it’s useful:

* [I need it](https://github.com/klmr/modules/blob/44db137fbdb431bc0f75422dfc239db1e7214f2e/R/help.r#L2) for ‹[modules](/klmr/modules)› (see also the above-mentioned issue)
* [Others need it](https://cran.r-project.org/web/packages/document/index.html).

I appreciate that “this is currently outside the scope of roxygen” as Hadley said. However, I think it’s a worthwhile addition because:

1. It’s trivial to add and maintain (the current PR is a single function).
2. The alternative would be for me or somebody else to create an exact copy of about 90% of roxygen2, maintained as a separate code base and separate CRAN project.
3. It’s useful: the output of the function ([after applying roclets](https://github.com/klmr/modules/blob/44db137fbdb431bc0f75422dfc239db1e7214f2e/R/help.r#L13-L14), which are already part of the public API of roxygen2) can be fed to `tools::parse_Rd` and then to e.g. `tools::Rd2txt`, and displayed. In fact, [that’s what ‹modules› is doing](https://github.com/klmr/modules/blob/44db137fbdb431bc0f75422dfc239db1e7214f2e/R/display_help.r). If there’s demand, it would be trivial to refactor the “display help from Rd” functionality into a separate package.

### DIFF
--- a/R/parse.R
+++ b/R/parse.R
@@ -29,14 +29,12 @@ parse_package <- function(base_path, load_code, registry, global_options = list(
 #' @return A list of roxygen2 blocks.
 #' @export
 #' @keywords internal
-parse_code <- function(
-  file, text = NULL,
-  env,
-  registry = default_tags(),
-  wrap = FALSE,
-  markdown = markdown_global_default,
-  file_encoding = "UTF-8"
-) {
+parse_code <- function(file, text = NULL,
+                       env,
+                       registry = default_tags(),
+                       wrap = FALSE,
+                       markdown = markdown_global_default,
+                       file_encoding = "UTF-8") {
   if (missing(file) && !is.null(text)) {
     file <- tempfile()
     writeLines(text, file)
--- a/R/parse.R
+++ b/R/parse.R
@@ -10,6 +10,38 @@ parse_package <- function(base_path, load_code, registry, global_options = list(
   list(env = env, blocks = blocks)
 }
 
+#' Parse roxygen2 comments in a piece of code.
+#'
+#' [parse_code()] takes a file or character vector, and an environment with R
+#' values, and parses the roxygen2 comments found in it.
+#'
+#' @param file Path of the file to process.
+#' @param env An environment containing the result of parsing and evaluating
+#'   the file.
+#' @param registry A roclet tag registry.
+#' @param wrap Logical; should roxygen2 output be wrapped? `FALSE` by default.
+#' @param markdown Logical value indicating whether to parse Markdown tags.
+#'   This can be overridden locally by the tags '@rd` and `@NoRd` inside a
+#'   Roxygen comment.
+#' @param file_encoding The file encoding. Default: `"UTF-8"`.
+#' @param text Instead of specifying `file`, users can also specify a character
+#' vector `text`, containing the R code.
+#' @return A list of roxygen2 blocks.
+#' @export
+parse_code <- function(file, env, registry = default_tags(), wrap = FALSE,
+                       markdown = markdown_global_default,
+                       file_encoding = "UTF-8", text) {
+  if (missing(file) && ! missing(text)) {
+    file <- tempfile()
+    writeLines(text, file)
+    on.exit(unlink(file))
+  }
+
+  options <- list(wrap = wrap, markdown = markdown)
+  parse_blocks(file, env, registry = registry, global_options = options,
+               fileEncoding = file_encoding)
+}
+
 parse_text <- function(text, registry = default_tags(), global_options = list()) {
   file <- tempfile()
   writeLines(text, file)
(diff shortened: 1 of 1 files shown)


hadley:
The major downside of this is that it exposes an internal data structure, preventing me from making changes in the future.

klmr:
Yes, and I would be more than happy to wrap this in some way.

*However*, in a way the structure is already part of the public API because it’s the input to the (public but mostly undocumented) `roclet_process` generic.

R/parse.R
```diff
@@ -10,6 +10,38 @@ parse_package <- function(base_path, load_code, registry, global_options = list(
...
+#' @param env An environment containing the result of parsing and evaluating
+#'   the file.
+#' @param registry A roclet tag registry.
+#' @param wrap Logical; should roxygen2 output be wrapped? `FALSE` by default.
+#' @param markdown Logical value indicating whether to parse Markdown tags.
+#'   This can be overridden locally by the tags '@rd` and `@NoRd` inside a
+#'   Roxygen comment.
+#' @param file_encoding The file encoding. Default: `"UTF-8"`.
+#' @param text Instead of specifying `file`, users can also specify a character
+#' vector `text`, containing the R code.
```
hadley: Needs indent.  Maybe document with text?

Can you please add `@keyword internal`

R/parse.R
```diff
@@ -10,6 +10,38 @@ parse_package <- function(base_path, load_code, registry, global_options = list(
...
+                       markdown = markdown_global_default,
+                       file_encoding = "UTF-8", text) {
+  if (missing(file) && ! missing(text)) {
+    file <- tempfile()
+    writeLines(text, file)
+    on.exit(unlink(file))
+  }
+
+  options <- list(wrap = wrap, markdown = markdown)
+  parse_blocks(file, env, registry = registry, global_options = options,
```
hadley: Can you please follow new standard for wrapping long lines? (http://style.tidyverse.org/syntax.html#long-lines)

R/parse.R
```diff
@@ -10,6 +10,38 @@ parse_package <- function(base_path, load_code, registry, global_options = list(
...
+#'   Roxygen comment.
+#' @param file_encoding The file encoding. Default: `"UTF-8"`.
+#' @param text Instead of specifying `file`, users can also specify a character
+#' vector `text`, containing the R code.
+#' @return A list of roxygen2 blocks.
+#' @export
+parse_code <- function(file, env, registry = default_tags(), wrap = FALSE,
+                       markdown = markdown_global_default,
+                       file_encoding = "UTF-8", text) {
+  if (missing(file) && ! missing(text)) {
```
hadley: I don't love interfaces where you have to specify exactly one of a set of arguments. I don't see any alternative here, but I think you need to make the structure clearer by placing `file` and `text` close together in the argument list. It might be better to default `text = NULL` and then condition on that.

R/parse.R
klmr: Did you mean “… document with *`file`*”? Otherwise I’m not sure what you mean by this.

R/parse.R
klmr: Done, here. Does this also apply to the parameter list in the function definition?

R/parse.R
hadley: Yes, sorry.

R/parse.R
hadley: Yup

klmr:
Please take another look.

R/parse.R
```diff
@@ -24,22 +26,31 @@ parse_package <- function(base_path, load_code, registry, global_options = list(
...
-#' vector `text`, containing the R code.
 #' @return A list of roxygen2 blocks.
 #' @export
-parse_code <- function(file, env, registry = default_tags(), wrap = FALSE,
-                       markdown = markdown_global_default,
-                       file_encoding = "UTF-8", text) {
-  if (missing(file) && ! missing(text)) {
+#' @keywords internal
+parse_code <- function(
+  file, text = NULL,
```
hadley: Oh except the rule for function blocks is that you indent with the `(` in function. I need to document that in style.

klmr: Ah yes, I found an example elsewhere in the code base. Makes sense.

hadley:
Thanks!

Issue #631, opened 2017-05-22
Possible regression in support of setMethodS3() and setConstructorS3() of R.methodsS3 and R.oo


Lucaweihs:
#525 added support for `R.methodsS3::setMethodS3(name, class, definition)` and `R.oo::setConstructorS3(name, definition)`. With the newest version of Roxygen, attempting to run roxygen using these methods results in an

```
Error in methods::is(x, "refObjectGenerator") : object 'value' not found
```

error for me. I've created a minimal package demonstrating the error [here](https://github.com/Lucaweihs/TestingRoxygen). The error goes away when prepending the package namespace to `setConstructorS3` and `setMethodS3` (that is, always using the fully namespaced `R.oo::setConstructorS3` and `R.methodsS3::setMethodS3`) but, when doing so, the s3 methods are not appropriately registered by the namespace (they are simply exported).

Lucaweihs:
@HenrikBengtsson

PR #525, opened 2016-10-10
Support setMethodS3() and setConstructorS3() of R.methodsS3 and R.oo

HenrikBengtsson:
`R.methodsS3::setMethodS3(name, class, definition)` defines S3 method `name.class <- definition` and `R.oo::setConstructorS3(name, definition)` defines R function `name <- definition`.  With this patch, roxygen2 recognizes the above constructors accordingly.

For the records, the background for this PR is private email 'R.methodsS3 with roxygen2' on 2016-10-03 from @martynplummer to me and @hadley (cc: @Lucaweihs and Mathias Drton).  The patch has been validated by me and independently by @martynplummer.

PR #639, opened 2017-07-14
Fixing a bug that broke parsing of R.methodsS3::setMethodS3 and R.oo:::setConstructorS3.

Lucaweihs:
The fixed bug was introduced in code review of pull request #525 (the arguments to a call to standardise_obj were incorrectly changed). This also fixes issue [631](https://github.com/klutometis/roxygen/issues/631).


hadley:
Do you think you could also include a small test case?

hadley:
Thanks!

PR #646, opened 2017-08-05
Modify ignore_files() for empty .Rbuildignore

dpritchLibre:
If the `.Rbuildignore` file is empty, or more generally if all of the lines in the
file are empty, then the `ignore_files` function in [utils.R](https://github.com/klutometis/roxygen/blob/master/R/utils.R) (line 106) will
throw an error rather than simply returning its first argument to the function,
`rfiles`, which is the desired behavior.  In more detail, if the `.Rbuildignore`
file is empty, then on line 117

    patterns <- patterns[patterns != ""]

returns a length-0 vector, which in turn causes

    matches <- lapply(patterns, grepl, rfiles_relative, perl = TRUE)

to return a length-0 list, which in turn causes

    matches <- Reduce("|", matches)

to return `NULL`.  This causes the !matches expression in

    rfiles[!matches]

to throw an error.  The proposed modification simply checks whether
`patterns[patterns != ""]` is a length-0 vector, and if so returns the first
argument to the function, `rfiles`.


R/utils.R
```diff
@@ -115,9 +115,14 @@ ignore_files <- function(rfiles, path) {
   # Remove any files that match any perl-compatible regexp
   patterns <- readLines(rbuildignore, warn = FALSE)
   patterns <- patterns[patterns != ""]
-  matches <- lapply(patterns, grepl, rfiles_relative, perl = TRUE)
-  matches <- Reduce("|", matches)
-  rfiles[!matches]
+  if (length(patterns) == 0L) {
+    rfiles
```
hadley: This should be `return(files)` so you don't need the `else` branch

dpritchLibre:
Revised the modification to incorporate the suggestion in the above comment.

Note that this pull request is a suggested fix to issue #647.

hadley:
Thanks - can you please add a bullet to news (matching the existing style), and make sure that you include `Closes #647` in the commit message?

hadley:
Thanks!

Issue #359, opened 2015-07-16
'dontrun' example file


cdeterman:
Is there any way to use the `@example` and `\dontrun`?  It is quite simple to use the normal `@examples` and `\dontrun` but the examples I work with can be more verbose and I don't want it cluttering my code.  They may also be used in other functions documentation.  However, I don't want the examples to be run.  Is there a way to accomplish this?

test.R
```r
# this does not work
#' @title test
#' @example \dontrun{examples/test_example.R}
test <- function(){
    print("hello")
}
```
```r
# this does
#' @title test
#' @examples 
#' \dontrun{
#' test()
#' }
test <- function(){
    print("hello")
}
```

test_example.R
```r
test()
```

P.S.  I posted this question of [SO](http://stackoverflow.com/questions/31412724/roxygen2-how-to-not-run-example-file) but it wasn't receiving much attention.

hadley:
No, sorry. You'd have to put the `\dontrun{}` inside `test_example.R`

PR #687, opened 2017-12-07
Added tags to wrap examples written in stand alone syntactically valid R scripts within the \dontrun, \dontshow, or \donttest environments.

dewittpe:
This feature address klutometis/roxygen#359

As part of my development process I frequently like to place examples in stand alone files for incremental testing and eventual documentation.  As such, being able to `source` the file is extremely helpful.  When the `\dontrun`, `\dontshow`, or `\donttest` wrappers are needed and added to the stand alone file such that the documentation is correct when using the `@example` tag the file can no longer be `source`d.

My solution: add three new tags, `@exampleDontrun`, `@exampleDontshow`, and `@exampleDonttest`.  These tags work the same way that `@example` does but wraps the syntactically valid R scripts in the respective environment.

Additional note: this feature has been noted at least twice on SO,
* https://stackoverflow.com/questions/12038160/how-to-not-run-an-example-using-roxygen2
* https://stackoverflow.com/questions/31412724/roxygen2-how-to-not-run-example-file

### DIFF
--- a/R/rd.R
+++ b/R/rd.R
@@ -34,6 +34,9 @@ roclet_tags.roclet_rd <- function(x) {
     encoding = tag_value,
     evalRd = tag_code,
     example = tag_value,
+    exampleDontrun = tag_value,
+    exampleDontshow = tag_value,
+    exampleDonttest = tag_value,
     examples = tag_examples,
     family = tag_value,
     field = tag_name_description,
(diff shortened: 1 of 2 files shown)


romanzenka:
I ran into this problem - if you put \dontrun{ } inside the example file, it is no longer valid R code, so will be flagged by static code analysis.

Your solution solves this. Any eta on having this added to Roxygen2?

hadley:
I think this would need to be tackled as part of #868, otherwise the order of the components will be lost.

PR #694, opened 2017-12-28
Fix docs re: auto-hyperlinking of URLs

### LINKED ISSUE #693: URLs are not automatically hyperlinked
The markdown vignette claims that "URLs are also automatically converted to hyperlinks" but I don't find that to be true.

https://github.com/klutometis/roxygen/blame/e2ed231a17644fb85b73a1d31bb1c84cb3bb7893/vignettes/markdown.Rmd#L138-L142

In my hands, I still need to surround with angle brackets.

jennybc:
Fixes #693

### DIFF
--- a/vignettes/markdown.Rmd
+++ b/vignettes/markdown.Rmd
@@ -135,10 +135,10 @@ Markdown hyperlinks work as usual:
 #' [Commonmark web site](http://commonmark.org/help)
 ```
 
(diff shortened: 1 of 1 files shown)


tbates:
makes sense to me: just notes users must enclose links in angle brackets if they want them to be auto-magically hyperlinked

hadley:
Thanks!

Issue #627, opened 2017-05-03
"String must be an atomic vector" thrown after upgrade to R 3.4.0


jonlehrer:
I recently upgraded to R 3.4.0. Packages that previously documented fine, now throw the following error:
"Error: String must be an atomic vector"
If I create an empty package and add a roxygen2 documentation block for the hello function, the same error occurs:

Build Messages:
==> devtools::document(roclets=c('rd', 'collate', 'namespace'))

Updating deleteme documentation
Loading deleteme
First time using roxygen2. Upgrading automatically...
Warning: @return [hello.R#19]: requires a value
Warning: @examples [hello.R#22]: requires a value
Error: String must be an atomic vector
Execution halted

Exited with status 1.

Session Info:
> sessionInfo()
R version 3.4.0 (2017-04-21)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 14.04.5 LTS

Matrix products: default
BLAS: /usr/lib/libblas/libblas.so.3.0
LAPACK: /usr/lib/lapack/liblapack.so.3.0

locale:
 [1] LC_CTYPE=en_CA.UTF-8       LC_NUMERIC=C               LC_TIME=en_CA.UTF-8        LC_COLLATE=en_CA.UTF-8    
 [5] LC_MONETARY=en_CA.UTF-8    LC_MESSAGES=en_CA.UTF-8    LC_PAPER=en_CA.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_CA.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] deleteme_0.1.0

loaded via a namespace (and not attached):
[1] compiler_3.4.0 tools_3.4.0

FvD:
I have the exact same issue but only in Ubuntu 16.04. On windows after the same upgrade to R 3.4 and on trying to build the documentation for the same package, there is no issue.

FvD:
@jonlehrer the issue dissappeared on my system after installing the package `lintr` by [jimhester](https://github.com/jimhester/lintr). I do not know what `roxygen` dependency was solved by installing all `lintr` dependencies, but hopefully this solves the issue for you as well.

florianm:
@FvD I have the same bug, your solution works, thanks!

[...]

PR #703, opened 2018-01-26
Update DESCRIPTION: Dependence for stringr changed from 0.5 to 1.0.0

tpbilton:
I had an issue related to this [issue](https://github.com/klutometis/roxygen/issues/627) and discovered that the dependence that was causing the problem was stringr. Thus, I have updated the DESCRIPTION file accordingly.

There was a change in the str_trim() function in the stringr package from version 0.6.2 to 1.0.0. The old version of str_trim() was causing problems with roxygenise().


hadley:
Thanks!

PR #719, opened 2018-02-28
Set width argument of wrapString() call to 75 for usage section

JoshOBrien:
At present, it's set to 80. Given that a further 5 space indent is added to the Usage section in the process of compiling rendered help files from Rd files, this means that the section can have width up to 85. This causes unattractive/distracting wrap around of some lines in textual help output (i.e. that got by doing `help(., help_type="text")`) in devices whose default viewing window is 80 characters wide.

Not sure that my proposed solution is optimal, but it involves few enough lines of code that it should be easy to review.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,8 +55,13 @@
 
 ## Minor improvements and bug fixes
 
-* `person` now supports all [MARC Relator](http://www.loc.gov/marc/relators/relaterm.html)
-  role codes (#662, @publicus).
+* `topic_add_usage()` now outputs formatted "Usage" section with max
+  width of 80 characters thanks to a now more flexible `wrap_string()`
+  (@JoshOBrien, #719).
+	
+* `person()` now supports all
+  [MARC Relator](http://www.loc.gov/marc/relators/relaterm.html) role codes
+  (#662, @publicus).
 
 * Empty `.Rbuildignore` now handled correctly (#576).
 
--- a/R/rd.R
+++ b/R/rd.R
@@ -331,7 +331,7 @@ topic_add_keyword <- function(topic, block) {
 # Prefer explicit \code{@@usage} to a \code{@@formals} list.
 topic_add_usage <- function(topic, block) {
   if (is.null(block$usage)) {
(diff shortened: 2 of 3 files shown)


hadley:
I like the approach. 

Can you please figure out why travis is failing?

JoshOBrien:
@hadley I took a look and figured out the issue. Travis failed because some of the testthat tests failed, specifically those in which `roclet_process` got called on a roclet without a usage section. When that happens, `topic_add_usage(rd, block)` gets called with `rd=NULL`, which ends up dispatching to `wrap.string.NULL()`. In my earlier pull request, I had neglected to add a `width=` argument to   `wrap.string.NULL, so when passed a width argument, it failed with a complaint about an unused argument.

The solution is to amend my earlier pull request with one additional edit, changing the definition of  `wrap.string.NULL` from:

    wrap_string.NULL <- function(x) return(x)

to something like

    wrap_string.NULL <- function(x, width = NULL) return(x)

Since I don't know the best Git way to add that patch to my earlier pull request, I'll let you deal with it by accepting the earlier pull request and then adding a fix like the one above with a second commit by hand, OK?

hadley:
If you want to try making the change, you can just commit to the branch you made the changes from. 

Otherwise, can you please enable edit access to your PR for me?  See https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/#enabling-repository-maintainer-permissions-on-existing-pull-requests for details.

JoshOBrien:
@hadley Hi -- gave it a shot and it looks from my end like my second commit got to you and passed all checks. Please let me know if I'm seeing that wrong or you need something else from me.

Thanks!

hadley:
FWIW, it's a bit cleaner to provide `width` as an argument to the generic, rather than `...`. But I'll make that change locally.

To finish the PR off, can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

JoshOBrien:
@hadley Done. Thanks.

hadley:
Thanks!

PR #731, opened 2018-04-13
#Always use UTF-8 encoding when parsing DESCRIPTION files

### LINKED ISSUE #727: Encoding issue in package doc when accented name
I am using the latest (dev) version that I need because of the new MARC roles. When using the package-level doc, the accent in my name can't be rendered.

Steps

* `usethis::create_package("test")`

* modify "First" to "Maëlle"

```
Package: test
Version: 0.0.0.9000
Title: What the Package Does (One Line, Title Case)
Description: What the package does (one paragraph).
Authors@R: person("Maëlle", "Last", email = "first.last@example.com", role = c("aut", "cre"))
License: What license is it under?
Encoding: UTF-8
LazyData: true
ByteCompile: true
RoxygenNote: 6.0.1.9000

```

* `usethis::use_package_doc()`

* `roxygen2::roxygenise()`

Resulting test-package.Rd

```
% Generated by roxygen2: do not edit by hand
% Please edit documentation in R/test-package.R
\docType{package}
\name{test-package}
\alias{test}
\alias{test-package}
\title{test: What the Package Does (One Line, Title Case)}
\description{
What the package does (one paragraph).
}
\author{
\strong{Maintainer}: MaÃ«lle Last \email{first.last@example.com}

}
\keyword{internal}
````

jimhester:
Fixes #727

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -149,10 +149,10 @@ block_eval <- function(tag, block, env, tag_name) {
 
 # Parse DESCRIPTION into convenient format
 read.description <- function(file) {
-  dcf <- read.dcf(file, keep.white = "Authors@R")
+  dcf <- desc::desc(file = file)
 
-  dcf_list <- setNames(as.list(dcf[1, ]), colnames(dcf))
-  lapply(dcf_list, str_trim)
+  fields <- dcf$fields()
+  purrr::map(purrr::set_names(fields), ~ dcf$get_field(.x))
 }
 
 
--- a/tests/testthat/test-rd-package.R
+++ b/tests/testthat/test-rd-package.R
@@ -18,3 +18,7 @@ test_that("can create package documentation", {
   expect_equal(get_tag(out, "docType")$values, "package")
   expect_equal(get_tag(out, "details")$values, "Details.")
 })
+
+test_that("Can read UTF-8 DESCRIPTIONS", {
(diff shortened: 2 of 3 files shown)


gaborcsardi:
Well, ideally we should read it in the specified encoding, I think.

Actually, I think we could replace `read.dcf` with the desc package, we are already using it apparently. That (well, the dev version) handles encodings correctly AFAIR.

jimhester:
Oh we are using desc? Then yes we should use it I will update this

hadley:
Do you want to finish this off @jimhester?

jimhester:
Ok I switched to using desc for this, however I am not sure if there needs to be additional changes as a result, the tests pass but there maybe some code paths that are untested which could fail.

gaborcsardi:
And these as well?
https://github.com/klutometis/roxygen/blob/651a8db990213ae67b22f0de1b4d6cf74b0e0c5b/R/object-defaults.R#L32-L44

`desc::desc()` returns  an R6 object, not just a list, so it is not a drop in replacement.

hadley:
Hmmm, maybe it's better to just fix the one `read.dcf()` call in roxygen2?  (And consider fuller desc usage at a later date)

gaborcsardi:
Or just just fix the`read.desciption`call to create a list from the R6 object, and then the rest stays the  same.

jimhester:
Ok I switched this to have `read.description()` just create a named list from the R6 object and we can handle a more thorough use of `desc` in the future.

hadley:
Thanks!

PR #732, opened 2018-04-20
Display name of help page from which section couldn't be inherited

krlmlr:
If an inherited section cannot be found, the warning contains the help  page from which that section was requested.


hadley:
Thanks!

PR #736, opened 2018-05-07
Skip devtools tests if unavailable.

QuLogic:
There's a (Suggests) dependency loop here. As there's only two uses here, it's pretty simple to add the skip for them in this package.

### DIFF
--- a/tests/testthat/test-rd-data.R
+++ b/tests/testthat/test-rd-data.R
@@ -1,6 +1,8 @@
 context("Rd: data")
 
 test_that("can document eager data", {
+  skip_if_not_installed("devtools")
+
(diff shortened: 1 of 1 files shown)


hadley:
A suggested dependency loop isn't a problem.

QuLogic:
It does cause a minor inconvenience when packaging. During initial bootstrap, only one or the other is available. Of course, it's only my personal preference that tests run during bootstrap, so if you don't want to apply this, that's fine.

hadley:
Ah ok!

hadley:
Thanks!

QuLogic:
Thanks!

Issue #745, opened 2018-06-19
Request: pkgdown site


alexpghayes:
It's now my go-to way to read documentation, and I suspect it may be for others as well.

hadley:
https://roxygen2.r-lib.org 🎉

PR #746, opened 2018-06-22
Added pkgdown site to Roxygen2.

md0u80c9:
Addresses issue #745 by creating pkgdown documentation.


hadley:
I'd rather do this myself — we also need to figure out some hosting issues.

PR #752, opened 2018-06-29
Propose fix some typos

jeis2497052:
here a couple of spelling corrections if that is still OK?  One file had some trailing white space ?


hadley:
Thanks!

PR #756, opened 2018-07-09
Stop roxygenize() quickly if DESCRIPTION not found (#704)

### LINKED ISSUE #704: Better error message when running roxygenize in the wrong dir
Following this [conversation on Twitter](https://twitter.com/hadleywickham/status/956876554512097280) :

I think the error message we get when trying to roxygenize() in a dir which is not the package root dir could be improved. For now, it's `Error: File file does not exist`

```
> getwd()
[1] "/Users/colin/Seafile/documents_colin/R/attempt"
> # Setting a path that is not the pkg root
> setwd("R/")
> getwd()
[1] "/Users/colin/Seafile/documents_colin/R/attempt/R"
> 
> roxygen2::roxygenise()
First time using roxygen2. Upgrading automatically...
Erreur : File file does not exist
> # Setting the wd to the package root
> setwd("..")
> getwd()
[1] "/Users/colin/Seafile/documents_colin/R/attempt"
# No pblm :
> roxygen2::roxygenise()
```
Maybe something like `Are you calling roxygenise() in a directory which is not the package root directory?`. 

Colin

mikmart:
I added an early check to `first_time()` (which runs at at the start of `roxygenize()`) to see if a DESCRIPTION file exists in the given folder. If not, stop with an informative error message. This fixes #704.

Further, previously a `man` directory was already created in the `roxygenize()`'d folder before an attempt to update the DESCRIPTION file in `roxygen_setup()` failed with an uninformative "file not found" error if no DESCRIPTION file was present in the directory. Now the function is stopped before that happens.

I feel like this could use a test somewhere, but skimming through the existing test, I wasn't sure where to fit one in. I was thinking about a test to assert that if `roxygenize()` can't find the DESCRIPTION file, it stops with a useful error message and without side effects (such as the added `man` folder).


hadley:
A test would be great - if should go in a test file called `test-safety.R`

R/safety.R
```diff
@@ -1,4 +1,11 @@
 first_time <- function(path) {
+  description <- file.path(path, "DESCRIPTION")
+  if (!file.exists(description)) {
+    stop("Could not find a DESCRIPTION file in ", path, ". ",
```
hadley: Per http://style.tidyverse.org/error-messages.html#problem-statement, can you please reword to use "must"?

mikmart: Yup! Updated the message.

mikmart:
Just rebased this, @hadley do you think the error message is OK now, or do you think this still needs adjusting?

hadley:
Looks good - thanks!

PR #763, opened 2018-07-26
Add missing backticks around `NAMESPACE` in NEWS

batpigandme:
This PR adds missing backticks around `NAMESPACE` where referenced without them in NEWS.md.


hadley:
Thanks!

PR #782, opened 2018-08-13
Add version requirement for `desc` package. Fixes #779.

### LINKED ISSUE #779: Error with a brand new project
After updating roxygen2 to 6.1.0 it keeps showing the following error, even with the brand new projects set up by RStudio (1.1.456) defaults, i.e. containing a single hello.R file.

```
First time using roxygen2. Upgrading automatically...
Error in .f(.x[[i]], ...) : attempt to apply non-function
Calls: suppressPackageStartupMessages ... read_pkg_description -> read.description -> <Anonymous> -> .Call -> .f
Execution halted

Exited with status 1.
```

rubak:
The function `get_field()` used in `read.description()` in the file `R/utils.R` was introduced in version 1.2.0 of the `desc` package. A system with older versions of `desc` fails with the error message reported in issue #779 which is rather cryptic.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,7 @@ Depends:
 Imports: 
     brew,
     commonmark,
-    desc,
+    desc (>= 1.2.0),
     digest,
     methods,
     pkgload,


gaborcsardi:
LGTM!

jon-harmon-macmillan:
Duplicates PR #773

rubak:
Indeed a duplicate. I didn't spot #773 since it doesn't appear under issues, and I didn't think about checking pull requests. Sorry for the noise. Should I just close this PR?

PR #800, opened 2018-10-09
Old version installed from Github

eribul:


gaborcsardi:
The  `klutometis/roxygen` repo actually has roxygen2:
https://github.com/klutometis/roxygen/blob/master/DESCRIPTION#L1

eribul:
Ah, ok :-)

PR #811, opened 2018-10-23
Fix MD link test case

### LINKED ISSUE #809: Unit test failure after commonmark 1.6 upgrade
In the latest version of commonmark on CRAN, text nodes in the ouptput from `commonmark_xml` now include attribute `xml:space="preserve"` to hint at libxml to not trim whitespace.

Unfortunately this breaks one of the unit tests in roxygen2, which uses a regex to compare output.

Before:

```xml
  <paragraph>
    <text>foo </text>
    <link destination="R:func()" title="">
      <text>func()</text>
    </link>
    <text> bar</text>
  </paragraph>
```

Is now:

```xml
  <paragraph>
    <text xml:space="preserve">foo </text>
    <link destination="R:func()" title="">
      <text xml:space="preserve">func()</text>
    </link>
    <text xml:space="preserve"> bar</text>
  </paragraph>
```

@gaborcsardi  @jennybc @kivikakk

gaborcsardi:
Closes #809. Do not regex match XML, that was a bad idea.

### DIFF
--- a/tests/testthat/test-rd-markdown-links.R
+++ b/tests/testthat/test-rd-markdown-links.R
@@ -71,33 +71,25 @@ test_that("% in links are escaped", {
 
 test_that("commonmark picks up the various link references", {
   cases <- list(
-    c("foo [func()] bar",
-      "<link destination=\"R:func\\(\\)\" title=\"\">\\s*<text>func\\(\\)</text>"),
-    c("foo [obj] bar",
-      "<link destination=\"R:obj\" title=\"\">\\s*<text>obj</text>"),
-    c("foo [text][func()] bar",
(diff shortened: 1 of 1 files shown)


jeroen:
Thanks! Sorry for not catching this during the commonmark release.

hadley:
Thanks - I'll try and get a release out later today

gaborcsardi:
@hadley oh, this is just in the tests, nothing is broken in roxygen itself. So just for this it is not urgent.

jennybc:
I have something else I'd like to track down (which is how I came across this). If CRAN isn't bothered yet, maybe a release could wait a bit?

hadley:
There are other problems which mean I'd rather get a release out sooner rather than later. I'm also happy to look into your problem.

PR #812, opened 2018-10-23
Fix warnings in collate tests

gaborcsardi:
Getting the `Collate` now checks that the files actually
exist, and warns if they don't. So we add dummy files to
the test case.

Fixes issue reported here:
https://github.com/klutometis/roxygen/commit/cc34200a598d02850020c874fdd0d6a82747f28e#commitcomment-31002481


hadley:
Thanks!

PR #814, opened 2018-10-24
Improve the documentation of links to help topics

### LINKED ISSUE #813: Unbalanced backticks in Rd for a [link text][pkg::object]-style link
This backtick behaviour: 

```
methods::`structure-class`
```

is inspired by how I would get help, e.g. by `` ?methods::`structure-class` ``

But the trailing backtick goes missing in the Rd docs.

I note that I got unstuck by *not* using backticks around `structure-class`. But it feels like these backticks should be tolerated / not mangled, since everything else in your experience suggests they are necessary or at least a good idea.

``` r
library(roxygen2)

roc_proc_text(
  rd_roclet(), "
    #' Title
    #'
    #' Description, see [`matrix`][methods::`structure-class`].
    #' @md
    foo <- function() {}"
)[[1]]
#> % Please edit documentation in RtmpiHM2q1/file13dc854fab817
#> \name{foo}
#> \alias{foo}
#> \title{Title}
#> \usage{
#> foo()
#> }
#> \description{
#> Description, see \code{\link[methods:`structure-class]
#> {matrix}}.
#> }
```

<sup>Created on 2018-10-23 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

This Rd leads to the [following WARNING on travis](https://travis-ci.org/tidyverse/tibble/jobs/444840369#L1271-L1275) but not for me locally (macOS) and I have no idea why.

```
❯ checking Rd cross-references ... WARNING
  Missing link or links in documentation object 'as_tibble.Rd':
    ‘[methods:`matrix-class]{matrix}’
  
  See section 'Cross-references' in the 'Writing R Extensions' manual.
```

jennybc:
Closes #813 

See a rendered version here (look at the Links section):

<http://rpubs.com/jennybc/markdown-vignette>

### DIFF
--- a/vignettes/markdown.Rmd
+++ b/vignettes/markdown.Rmd
@@ -154,27 +154,26 @@ First we explore the simplest form: `[ref]`. The presence of trailing parenthese
   * The link text `func()` is automatically typeset as code.
   * The parentheses are stripped in the derived Rd link target.
 
-+--------------------+------------------------+--------------------------------+
-| `[ref]`\           | Links to help\         | Notes                          |
-|  examples          | topic for ...          |                                |
-+:===================+:=======================+:===============================+
-| `[func()]`\        | a function in same\    | Always typeset as code.\                   |
-| `[pkg::func()]`    | package or in `pkg`    | Produces Rd: `\code{\link[=func]{func()}`\ | 
-|                    |                        | or `\code{\link[pkg:func]{pkg::func()}}`   |
-+--------------------+------------------------+--------------------------------+
-| `[thing]`\         | a topic in same\       | Use for a topic that documents `NULL`\ |
-| `[pkg::thing]`     | package or in `pkg`    | and name is set via `@name`,\          |
-|                    |                        | e.g., a dataset or concept.\           | 
-|                    |                        | Not typeset as code.\                  | 
-|                    |                        | Produces Rd: `\link{thing}` or\        |
-|                    |                        | `\link[pkg:thing]{pkg::thing}`         |
-+--------------------+------------------------+--------------------------------+
-| ``[`thing`]``\     | a topic in same\       | Same as above, but explicit backticks\ |
-| ``[`pkg::thing`]`` | package or in `pkg`    | mean that it **is** typeset as code.\  |
-|                    |                        | Good for documenting a class.          |
-|                    |                        | Produces Rd: `\code{\link{thing}}` or\ |
-|                    |                        | `\code{\link[pkg:thing]{pkg::thing}`}  |
-+--------------------+------------------------+--------------------------------+
++--------------------+------------------------+---------------------------------------------+
+| `[ref]`\           | Links to help\         | Notes                                       |
+|  examples          | topic for ...          |                                             |
++:===================+:=======================+:============================================+
+| `[func()]`\        | a function in same\    | Always typeset as code.\                    |
+| `[pkg::func()]`    | package or in `pkg`    | Produces Rd: `\code{\link[=func]{func()}}`\ | 
+|                    |                        | or `\code{\link[pkg:func]{pkg::func()}}`    |
++--------------------+------------------------+---------------------------------------------+
+| `[thing]`\         | a topic in same\       | Use for a topic that documents `NULL` and name is set\ |
+| `[pkg::thing]`     | package or in `pkg`    | via `@name`, e.g., a dataset or concept.\              |
+|                    |                        | Not typeset as code.\                                  | 
+|                    |                        | Produces Rd: `\link{thing}` or\                        |
+|                    |                        | `\link[pkg:thing]{pkg::thing}`                         |
++--------------------+------------------------+---------------------------------------------+
+| ``[`thing`]``\     | a topic in same\       | Same as above, but explicit backticks\      |
+| ``[`pkg::thing`]`` | package or in `pkg`    | mean that it **is** typeset as code.\       |
+|                    |                        | Good for documenting a class.\              |
+|                    |                        | Produces Rd: `\code{\link{thing}}` or\      |
+|                    |                        | `\code{\link[pkg:thing]{pkg::thing}`}       |
++--------------------+------------------------+---------------------------------------------+
 
 Use the second form `[text][ref]` to link to the topic specified by `ref`, but with `text` as the link text.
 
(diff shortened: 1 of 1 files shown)


jennybc:
@gaborcsardi I don't have enough permission to request a review from you, but here you go.

jennybc:
This table looks pretty good when the vignette is compiled to PDF. The internal links also seem to work.

vignettes/markdown.Rmd
```diff
@@ -2,7 +2,14 @@
...
 date: "`r Sys.Date()`"
-output: rmarkdown::html_vignette
+output:
+  html_document:
+    fig_caption: false
+    toc: true
+    toc_float:
+      collapsed: false
+      smooth_scroll: false
+    toc_depth: 3
```
gaborcsardi: What's the benefit of `html_document` over `html_vignette`?

vignettes/markdown.Rmd
```diff
@@ -141,33 +148,52 @@ URLs inside angle brackets are also automatically converted to hyperlinks:
...
-```
-
-Or a shorthand notation can be used: `[abc-class]` is converted to `\linkS4class{abc}` and `[pkg::abc-class]` is converted to `\link[pkg:abc-class]{pkg::abc}`.
+Markdown notation can also be used to create links to other help topics. There are two basic forms:
+
+  * `[ref]`: The target topic and the link text are the same.
+  * `[text][ref]`: Link text differs from the target.
+
+with variations depending on whether `ref` is in the same package and whether the link is automatically typeset as code. Use backticks inside the square brackets to typeset the link text as code, when it would not be otherwise:
+
```
gaborcsardi: The rule is that function names are always typeset as code, i.e. anything that ends with `()`. Otherwise you might want to typeset as code, might not.  For `[text][ref]`maybe not, since the `text` is might not be an object name, just generic text. So I think ``[`text`][ref]`` below is a bit misleading. 

TBH I am not even sure if we should typeset class names as code, maybe.  The style guide does not say anything about that.  I often just write 
```r
#' ... a [processx:process] object ...
```
If I want this as code, I think that would be
```r
#' ... a [`processx:process`][processx::process] object ...
```
which is quite cumbersome, and if we decide that class names should be code, we should probably have some shorthand for this.

vignettes/markdown.Rmd
```diff
@@ -141,33 +148,52 @@ URLs inside angle brackets are also automatically converted to hyperlinks:
...
++-----------------------+------------------------+--------------------------+  
+| `[object]`\           | a topic in\            | Use for a topic where you documented\ |
+| ``[`object`]``        | same package           | `NULL` and set its name via `@name`,\ |
+|                       |                        | e.g., a class, dataset, or theme.\    | 
+|                       |                        | Bring your own backticks.\            |
+|                       |                        | In Rd: `\link{object}`                |
++-----------------------+------------------------+--------------------------+
+| `[text][func()]`\     | a function or topic\   | Use when you want to provide your\               |
+| `[text][object]`\     | in same package        | own link text.\                                  |
+| `[abc][abc-class]`    |                        | Bring your own backticks.\                       |
```
gaborcsardi: For `text`  you usually don't want backticks, that is just text. So maybe the `abc-class` case could  be on its own row?

vignettes/markdown.Rmd
```diff
@@ -141,33 +148,52 @@ URLs inside angle brackets are also automatically converted to hyperlinks:
...
+|                       |                        | In Rd: `\link[=func]{text}`,                     |
+|                       |                        | `\link[=obj]{text}`, or `\link[=abc-class]{abc}` |
++-----------------------+------------------------+--------------------------+
+| `[pkg::func()]`       | a function in `pkg`    | Always typeset as code.\                     |
+|                       |                        | In Rd: `\code{\link[pkg:func]{pkg::func()}}` |
++-----------------------+------------------------+--------------------------+
+| `[pkg::object]`       | a topic in `pkg`       | Use for a non-function object in `pkg`\ |
+|                       |                        | Bring your own backticks.\              |
```
gaborcsardi: I am not even sure if you can use backticks in this case. Can you?

vignettes/markdown.Rmd
```diff
@@ -141,33 +148,52 @@ URLs inside angle brackets are also automatically converted to hyperlinks:
...
+|                       |                        | In Rd: `\link[pkg:object]{pkg::object}` |
++-----------------------+------------------------+--------------------------+
+| `[text][pkg::func()]` | a function or topic\   | Use when you want to provide your\ |
+| `[text][pkg::object]` | in `pkg`               | own link text.\                    |
+|                       |                        | Bring your own backticks.\         |
```
gaborcsardi: Again, `text` is almost never code IMO.

vignettes/markdown.Rmd
jennybc: It appears to work, yes. I just tried.

vignettes/markdown.Rmd
jennybc: Yeah I am only pointing out that you must use the backticks if you want `\code{...}`. I guess that is confusing.

vignettes/markdown.Rmd
jennybc: > The rule is that function names are always typeset as code, i.e. anything that ends with ()

So is it based on the presence of `()` or actually determining if the target is a function? Now that I say this, I supposed it's probably just based on the `()`.

I got into the cumbersome situation you're talking about with `data.frame` and `data.frame()` in tibble. There *is* no topic for the `data.frame` class. So when you talk about the class, it seems you have to link to the `data.frame()` constructor function. But I think "data.frame" would look weird if it were *not* typeset as code.

vignettes/markdown.Rmd
gaborcsardi: Got it.

vignettes/markdown.Rmd
jennybc: Yeah but there *could* and arguably *should* be a topic called "data.frame-class", like [httr::Token-class](https://www.rdocumentation.org/packages/httr/versions/1.3.1/topics/Token-class), [Matrix-class](https://www.rdocumentation.org/packages/Matrix/versions/1.2-14/topics/Matrix-class), etc.

vignettes/markdown.Rmd
jennybc: I took the simpler approach and updated the Rpubs preview: http://rpubs.com/jennybc/432771

jennybc:
OK I split into a table for `[ref]` form and `[text][ref]` form and reorganized rows. Rpubs is updated.

I think we're down to what we say about S3 and S4 classes.

hadley:
Merging as it's useful now; feel free to add S3/S4 whenever.

PR #841, opened 2019-01-19
Force unix-style line endings even on Windows systems.

### LINKED ISSUE #840: Always use unix line endings in NAMESPACE
When I generate documentation on a Windows machine, the NAMESPACE is rewritten with CR/LF instead of LF, causing it to appear that every line of NAMESPACE has changed.

jonthegeek:
Closes #840 . @hadley @batpigandme should this count in Tidyverse Developer Day stats?


hadley:
Would you mind adding a unit test for this? @jimhester might have ideas if it's not obvious.

jonthegeek:
> Would you mind adding a unit test for this? @jimhester might have ideas if it's not obvious.

Will do! Tracking down @jimhester for some advice.

jonthegeek:
> The other tests in test-utf8.R are failing for me locally now. I'm trying to wrap my head around why that would be. It appears to be something causing roxygenize to always regenerate the special-character-containing files, but that's as far as I've gotten.

I've been unable to determine what's causing these tests to fail. @jimhester any ideas?

jonthegeek:
The tests fail *on Windows* specifically (I used my linux server to test). Continuing to investigate why.

jonthegeek:
The tests now pass! The unexported `same_contents` function (called by a method called by a function called by `roxygenize`) had a special case for Windows to work around this issue.

tests/testthat/test-utf8.R
```diff
@@ -33,3 +33,14 @@ test_that("unicode escapes are ok", {
   # Shouldn't change again
   expect_output(roxygenise(test_pkg, roclets = "rd"), NA)
 })
+
+test_that("write_lines writes unix-style line endings.", {
+  path <- test_path("escapes.Rd")
+  temp_filename <- tempfile()
+  old_binary <- readBin(path, "raw", n = 100L)
```
jimhester:
```suggestion
  old_binary <- readBin(path, "raw", n = file.info(path)$size)
```

tests/testthat/test-utf8.R
```diff
@@ -33,3 +33,14 @@ test_that("unicode escapes are ok", {
...
 })
+
+test_that("write_lines writes unix-style line endings.", {
+  path <- test_path("escapes.Rd")
+  temp_filename <- tempfile()
+  old_binary <- readBin(path, "raw", n = file.info(path)$size)
+  old_text <- read_lines(path)
+  write_lines(old_text, temp_filename)
+  on.exit(unlink(temp_filename), add = TRUE)
+  new_binary <- readBin(temp_filename, "raw", n = 100L)
```
jimhester: There is also one more place with the same pattern

```suggestion
  new_binary <- readBin(temp_filename, "raw", n = file.info(temp_filename)$size)
```

jonthegeek:
All tests still pass on Windows + Linux. I think this is good to go.

hadley:
I talked this over with @jennybc and I may have completely misdiagnosed the problem 😢 It might actually be a problem with your git settings — for most people on windows, the different line endings should be translated automatically.

jonthegeek:
> I talked this over with @jennybc and I may have completely misdiagnosed the problem 😢 It might actually be a problem with your git settings — for most people on windows, the different line endings should be translated automatically.

Interesting! I'll dig through Happy Git with R to see if I can find it. That said... this also fixes it (after I implemented this I used it to build some other docs and no longer got the wall of changes). https://knowyourmeme.com/memes/why-not-both-why-dont-we-have-both ?

jennybc:
It might still be worth attending to this in roxygen. I don't have a great grasp of all the ways line endings play out in files edited manually vs. written programmatically.

But yeah generally I think a Windows + Git user will lead a happier life if this is set up optimally across all repos and activities. We've found that the current default settings for Git for Windows make all of this "just work". Here's a relevant screenshot from installation:

autocrlf

More background on recommended config for `core.autocrlf`:

https://help.github.com/articles/dealing-with-line-endings/#platform-windows

jonthegeek:
I'm not sure how mine ever became incorrect. The majority of my work with git is actually via Rstudio Server on linux, so I only use the Windows side for occasional non-work-related projects (like Tidyverse Dev Day), but I'm pretty sure I used the default settings when I set things up on this machine. I submitted a PR on happygitwithr about this, too: https://github.com/jennybc/happy-git-with-r/pull/124

r2evans:
My vote is with @jennybc and @jonthegeek in that this is something `git` is not solving before the commit. That is, before the commit, git thinks there is a change, even if it is told to ignore LF/CRLF differences within the repo. The problem is that git utilities will *report* some form of difference.

One thought is that the code try to infer which is in use (LF or CRLF) and continue using that. When a file is written, is it written completely (and not just appended), so I don't think a file can turn "inconsistent" between changes. This could be done with with `file` (which will report CRLF), but if you want it to work without Rtools then you'll need a R-centric way of doing that. It's not hard, perhaps just

```r
grepl("\r", readChar("NAMESPACE", 1e4))
```

where 1e4 should be sufficiently high to ensure at least one line-ending. It is when previously generated by `roxygen2`. I'm curious what the longest `NAMESPACE` single line looks like, but it might be easy enough to assume something ridiculous like 1e4; this check is simple enough and reading in that many bytes is not slow.

gaborcsardi:
Can we keep the current line endings for a file if it already exists? And default to LF otherwise? This would make git happy as well?

gaborcsardi:
Yeah. And we can document that to canonise the line endings, just remove the files. And/or we can have a DESCRIPTION option to force LF.

gaborcsardi:
@jimhester how about choosing LF if there is an LF without a CR you in front of it? Ie. For mixed files we choose LF?

gaborcsardi:
Same logic, just grep for `[^\r]\n` to default to LF, no? Anyway, not terribly important.

gaborcsardi:
Well, the file with only `\n` is not the case we care much about.... but it works for everything else? Anyway, again, not terribly important....

jimhester:
News has been updated!

PR #846, opened 2019-02-21
Add column headers to \tabular{} generation in the Text Formatting vignette

jhchou:
Modify `\tabular{}` table generation in the text formatting vignette to include a bolded column name row and `#' ` prefixes for roxygen documentation.

Having column headers seems like a common enough need (and the extra line is easily deleted if not wanted), and not having to prepend all the lines with `#' ` before pasting into the documentation seemed worthwhile.

Apologies if I'm doing this all wrong -- I've never attempted anything collaborative in GitHub before, so this tiny change is my first ever attempt at a Pull Request.


hadley:
Thanks!

Issue #767, opened 2018-07-31
allow multiple @inheritDotParams


maxheld83:
thanks so much for `#' @inheritDotParams`; it's a fantastic addition.

Though I couldn't find documentation for this, it appears that you can currently use only *one* `#' @inheritDotParams` per documentation entry.

When, all other things equal, I try to add *two* such tags for the same function, I get this error message:

```
Writing NAMESPACE
Error in mapply(FUN = f, ..., SIMPLIFY = FALSE) : 
  zero-length inputs cannot be mixed with those of non-zero length
Calls: suppressPackageStartupMessages ... topics_process_inherit -> <Anonymous> -> fun -> Map -> mapply
```

Is this a known/intended limitation of `#' @inheritDotParams`?

----

My use case is something like this:

```
#' Do something
#' Does something
#' @param x [`character(1)`] very interesting text.
#' @inheritDotParams lower_level_fun_a -x
#' @inheritDotParams lower_level_fun_b -x
higher_level_fun(x, ...) {
  lower_level_fun_a(x, ...)
  lower_level_fun_b(x, ...)
}
#' Do a
#' Does a
#' @param foo a very fooish character
lower_level_fun_a(x, foo = "zap") {
  print(x, foo)
}
#' Do b
#' Does b
#' @param bar a very booish character
lower_level_fun_b(x, bar = "zong") {
  print(x, zong)
}
```

I understand that passing on `...` to *multiple* functions is a bit risky, because *if* arguments carry the same names, they get matched twice (intended here).

But if this seems safe to the dev, *multiple* `#' @inheritDotParams` could save a lot of typing.

pllittle:
@maxheld83 I've also encountered this issue after multiple attempts to generate Rd files from roxygen.  Allowing for multiple inheritDotParams would definitely save on typing time.

gustavdelius:
@halldc, given that you have that nice pull request #842 for improving inheritDocParams, perhaps you have an idea how to fix this issue?

gustavdelius:
This error was due to a simple cut-and-paste error in a single line in the code. I have submitted pull request #854 which fixes it.

PR #854, opened 2019-03-26
Fixes for @inheritDocParams

gustavdelius:
Fixing #767


gustavdelius:
Now also fixes #857

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # roxygen2 (development version)
 
+## Minor improvements and bug fixes
```
hadley: Can you please remove this heading? As noted in https://style.tidyverse.org/news.html#news-release, we only add the subheads just prior to release.

NEWS.md
```diff
@@ -1,5 +1,9 @@
...
+
+* Allow multiple @inheritDotParams (#767, @gustavdelius)
```
hadley: Can you please also mention #857?

R/field.R
```diff
@@ -37,6 +37,14 @@ merge.roxy_field <- function(x, y, ...) {
   roxy_field_simple(x$field, c(x$values, y$values))
 }
 
+#' @export
```
hadley: I think you forgot to re-document to update the NAMESPACE?

R/field.R
```diff
@@ -37,6 +37,14 @@ merge.roxy_field <- function(x, y, ...) {
...
+merge.roxy_field_param <- function(x, y, ...) {
+  stopifnot(identical(class(x), class(y)))
+  # When parameters appear in both x and y, keep values from y
```
hadley: I would've thought that the first definition would win, not the last — I think that would be more consistent with `@inheritParams`?

gustavdelius: @hadley , this order is not the order in which the tags appear in the roxygen comments, but it is the order in which the code executes, and this order is first `inherit_params()` and then `inherit_dot_params()`. So if `inherit_params()` has added "..." already, then we want `inherit_dot_params()` to overwrite this. I have added a comment to explain this.

NEWS.md
gustavdelius: Done

R/field.R
gustavdelius: Done now.

R/field.R
hadley: In that case, I don't think you should do it in this more general `merge()` method; I think you should just inline this code directly in `merge.roxy_field_inherit_dot_params()`.

gustavdelius: The way I think about it is that the params field should never have any duplicates. If the code tries to set the same param value twice, it should be the later one that overwrites the earlier one, because that is how assignments in code generally work. With the code written this way, if someone else in the future again writes code that merges stuff into the params field (as the author of inhert_dot_params() did), duplicates will automatically be avoided. 

Of course I would be happy to follow your recommendations, but I am not sure how to do that elegantly. The function merge.roxy_field_inherit_dot_params() is never called with the params field, so how can it remove the old "..." slot from it that inherit_params() had added.

hadley: I'm just saying that this code should go in `merge.roxy_field_inherit_dot_params()`; in complicated packages like roxygen2 you want to make the simplest possible change to avoid unintended consequences.

Or are you saying that both merge methods need to be modified in order for your approach to work?

If that's the case, I think we need to avoid the `...` duplication at some other place, as making the last parameter definition win seems like a major change to me.

gustavdelius: I agree with keeping bigfixes simple. I also agree with keeping bugfixes local. However in this case we could describe the bug more generally as: "the params field sometimes contains duplicates". I agree that currently the duplicate happens to always be "...", but any other duplicate would be just as bad. So I think it is appropriate to fix the bug in a way that will always avoid duplicates, i.e., by not allowing `merge.roxy_field_params()` to create duplicates. It makes the code more robust against possibly someone in the future again writing code that produces duplicates.

I think it is also the simplest way to fix the bug. Currently both `inherit_params()` and `inherit_dot_params()` independently of each other decide what they want to add to the params field and each calls `merge.roxy_field_params()` to do that (via the `add_simple_field()` method), one after the other. When `inherit_params()` is running, it does not know yet that `inherit_dot_params()` will run after it and will add "...", so it adds its own "..." value to the params field by calling `merge.roxy_field_params()`. When `inherit_dot_params()` runs afterwards, it needs to overwrite this with its own "..." value and it does that by also calling `merge.roxy_field_params()`. The only problem is that currently `merge.roxy_field_params()` creates a duplicate "...".

If we don't want `merge.roxy_field_params()` to always avoid duplicates, we could tell `inherit_dot_params()` to check in the `RoxyTopic` object whether it already has a "..." value for its params field and delete that value before calling `merge.roxy_field_params()`. That however seems a bit more hacky to me, especially as there is no "remove" method currently provided by the `roxy_field` class. However I could of course add one. Would you prefer that route?

Her is one other option: the `add_simple_field()` and `add_field()` methods already have an argument called `overwrite = FALSE` that will make it overwrite the existing field. How about giving it another argument `overwrite_value = FALSE` that it would pass on to `merge()` and that would then allow `merge()` to overwrite existing values in the field, rather than creating duplicates. Then `inherit_dot_params()` could call `add_simple_field()` with `overwrite_value = TRUE`.

hadley:
Thanks @gustavdelius!

PR #875, opened 2019-06-29
Update Rd vignette to refer to Markdown vignette

### LINKED ISSUE #871: `rd.Rmd` vignette references wrong vignette for Markdown documentation
While reading the documentation, I noticed that Line 108 of the [rd.Rmd vignette](https://github.com/klutometis/roxygen/blob/master/vignettes/rd.Rmd) refers to the [formatting vignette](https://github.com/klutometis/roxygen/blob/master/vignettes/formatting.Rmd) for discussion of Markdown formatting. However, the formatting vignette does not mention Markdown. Perhaps Line 108 should simply be updated to refer to the [Markdown vignette](https://github.com/klutometis/roxygen/blob/master/vignettes/markdown.Rmd) instead?

mikemc:
Fixes #871

### DIFF
--- a/vignettes/rd.Rmd
+++ b/vignettes/rd.Rmd
@@ -99,13 +99,15 @@ This introductory block is broken up as follows:
    provide any other important details of how the function operates. The 
    details are optional.
 
+Note the wrapping of the roxygen block. You should make sure that your comments are less than ~80 columns wide.
+
 Text within roxygen blocks can be formatted using `.Rd` syntax; see [formatting](formatting.html) for examples and [R extensions](https://cran.r-project.org/doc/manuals/R-exts.html#Marking-text) for full details. You can also use Markdown formatting within roxygen blocks, as long as you have the following in the `DESCRIPTION`:
 
 ```
 Roxygen: list(markdown = TRUE)
 ```
 
-We'll discuss that more in [Markdown](markdown.html). Also note the wrapping of the roxygen block. You should make sure that your comments are less than ~80 columns wide.
+We'll discuss that more in [Markdown](markdown.html).
 
 ## Object specifics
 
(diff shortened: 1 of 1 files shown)


vignettes/rd.Rmd
```diff
@@ -97,15 +97,15 @@ This introductory block is broken up as follows:
 *  The third and subsequent paragraphs go into the __details__: this is a
    (often long) section that comes after the argument description and should 
    provide any other important details of how the function operates. The 
-   details are optional..
+   details are optional.
 
-Note that you can use markdown formatting within roxygen blocks, as long as you have the following in the `DESCRIPTION`:
+Text within roxygen blocks can be formatted using `.Rd` syntax; see [formatting](formatting.html) for examples and [R extensions](https://cran.r-project.org/doc/manuals/R-exts.html#Marking-text) for full details. You can also use Markdown formatting within roxygen blocks, as long as you have the following in the `DESCRIPTION`:
```
gaborcsardi: Do you think we can put the markdown part in a different paragraph?

vignettes/rd.Rmd
```diff
@@ -97,15 +97,15 @@ This introductory block is broken up as follows:
...
 
 ```
 Roxygen: list(markdown = TRUE)
 ```
 
-We'll discuss that more in [formatting](formatting.html). Also note the wrapping of the roxygen block. You should make sure that your comments are less than ~80 columns wide.
+We'll discuss that more in [Markdown](markdown.html). Also note the wrapping of the roxygen block. You should make sure that your comments are less than ~80 columns wide.
```
gaborcsardi: I think from "Also note..." belongs to the Rd stuff more, so maybe move that up after "...full details."?

mikemc: I agree this should be moved up. How about having it before any discussion of formatting, and then having the Rd and markdown formatting in one "formatting" paragraph. See fixup commit 
98ed691

mikemc: perhaps we should also add "or put `@md` in the block. " at the beginning of line 108, following the `Roxygen: list(markdown = TRUE)` code block

hadley:
Thanks!

PR #877, opened 2019-07-03
Fix grammatical error

mikemc:


gaborcsardi:
Thanks much, this part is completely rewritten in https://github.com/klutometis/roxygen/pull/814, which we should have merged a long time ago.

PR #883, opened 2019-07-22
support custom family prefixes (closes #830)

### LINKED ISSUE #830: allow customization of '@family' prefix
When using e.g.

    #' @family foo

The documentation gets an entry in the seealso section as:

    Other foo: <...>

I'd like to be able to customize the prefix, so that I could just write e.g.

    #' @family Environment management

and just have a plain old:

    Environment management: <...>

ie, without the `Other:` prefix. Does that seem sensible?

tl;dr: "Other:" isn't always a great prefix and I'd like to customize that somehow.

kevinushey:
This PR implements a simple mechanism for providing custom roxygen metadata, and uses that for supplying a custom family prefix.

If a file at `man-roxygen/roxygen-meta.R` exists, it will be sourced when roclets are processed. It should return an R list of options. Currently, only `family.prefix` is understood; this object should itself be an R list mapping family names to custom prefixes.


kevinushey:
Looks like codecov is being mean to my PR (which is odd since the new code should all be tested)

R/rd.R
```diff
@@ -71,6 +71,10 @@ roclet_process.roclet_rd <- function(x,
                                      env,
                                      base_path,
                                      global_options = list()) {
+
+  # Load roxygen metadata
+  roxy_meta_load(base_path)
```
hadley: I think this should go in `roxygenize()` so called before every roclet?

R/meta.R
```diff
@@ -0,0 +1,39 @@
...
+
+roxy_meta_read <- function(name) {
+  .roxygen_meta[[name]]
+}
+
+roxy_meta_load <- function(base_path) {
+
+  meta_dir <- file.path(base_path, "man-roxygen")
+
+  pattern <- "^roxygen-meta[.][Rr]"
```
hadley: It seems weird to use `roxygen` twice here. How about `man/roxygen/options.R`? Can you check if R CMD check warns about such a file/directory? (i.e. do we need to build-ignore it too?)

R/meta.R
```diff
@@ -0,0 +1,39 @@
...
+  meta_files <- list.files(meta_dir, pattern = pattern, full.names = TRUE)
+  if (length(meta_files) == 0)
+    return(FALSE)
+
+  if (length(meta_files) > 1) {
+    warning("multiple roxygen-meta.R files found: using '",
```
hadley: This should just be an error, I think.

R/meta.R
```diff
@@ -0,0 +1,39 @@
...
+            basename(meta_files[[1]]),
+            "'")
+    meta_files <- meta_files[[1]]
+  }
+
+  result <- tryCatch(eval(parse(meta_files)), error = identity)
+  if (inherits(result, "error")) {
+    warning(result)
```
hadley: I think this should be an error too:

```R
abort("Failed to parse `man/roxygen/meta.R", parent = result)
```

R/meta.R
```diff
@@ -0,0 +1,39 @@
...
+    return(FALSE)
+  }
+
+  if (!is.list(result)) {
+    warning("evaluation of roxygen-meta.R did not return a list")
```
hadley: Similarly here

R/meta.R
```diff
@@ -0,0 +1,39 @@
...
+    return(FALSE)
+  }
+
+  rm(list = ls(envir = .roxygen_meta), envir = .roxygen_meta)
+  list2env(result, envir = .roxygen_meta)
```
hadley: Could you just do `.roxgen_meta <<- result` ? I don't think it has to be an environment?

R/rd-family.R
```diff
@@ -1,3 +1,15 @@
+topics_process_family_prefix <- function(family) {
+
+  # read custom family prefix from roxygen meta
+  opts <- roxy_meta_read("family.prefix")
```
hadley: To avoid future namespace clashes, I think this should be `rd_family_prefix`. 

But I'm not sure `prefix` is the right word? Maybe `title` would be better since you're looking up the title from the prefix?

R/rd-family.R
```diff
@@ -1,3 +1,15 @@
...
+  if (!is.null(opts[[family]]))
+    return(opts[[family]])
+
+  # default prefix
+  paste("Other ", family, ": ", sep = "")
```
hadley: Since you touched this, can you please switch to `paste0()`?

R/rd-family.R
```diff
@@ -1,3 +1,15 @@
+topics_process_family_prefix <- function(family) {
+
+  # read custom family prefix from roxygen meta
+  opts <- roxy_meta_read("family.prefix")
+  if (!is.null(opts[[family]]))
```
hadley: Where should we check that this is a list?

R/meta.R
```diff
@@ -0,0 +1,45 @@
...
+  if (inherits(parsed, "error")) {
+    message <- "could not parse 'man/roxygen/meta.R'"
+    rlang::abort(message, parent = parsed)
+  }
+
+  # TODO: appropriate evaluation environment?
+  envir <- globalenv()
+  result <- tryCatch(eval(parsed, envir = envir), error = identity)
+  if (inherits(result, "error")) {
+    message <- "could not evaluate 'man/roxygen/meta.R'"
```
hadley:
```suggestion
    message <- "Could not evaluate 'man/roxygen/meta.R'"
```

R/meta.R
```diff
@@ -0,0 +1,45 @@
...
+    rlang::abort(message, parent = result)
+  }
+
+  if (!is.list(result)) {
+    message <- "evaluation of 'man/roxygen/meta.R' did not return a list"
```
hadley:
```suggestion
    message <- "Evaluation of 'man/roxygen/meta.R' did not return a list"
```

tests/testthat/test-rd-family.R
```diff
@@ -95,3 +95,31 @@ test_that("family also included in concepts", {
...
+  owd <- setwd(tempdir())
+  on.exit(setwd(owd), add = TRUE)
+
+  dir.create("man/roxygen", recursive = TRUE)
+  write_lines(
+    "list(rd_family_title = list(a = 'Custom prefix: '))",
+    "man/roxygen/meta.R"
+  )
+
+  roxy_meta_load(getwd())
```
hadley: Maybe it would better to have `roxy_meta_set()` that you could pass a list? Then have a separate set of tests for `roxy_meta_load()`?

R/rd-family.R
kevinushey: I think it's okay for `rd_family_title` to just be a named 'something' -- doesn't necessarily have to be a list. (We just enforce that the metadata object itself is a list). We could be stricter if you prefer though.

kevinushey:
I get this complaint during `devtools::document()` if `.roxygen_meta` is a plain object as opposed to an environment:

```
Updating roxygen2 documentation
Error in .roxygen_meta <<- result : 
  cannot change value of locked binding for '.roxygen_meta'
Calls: suppressPackageStartupMessages ... <Anonymous> -> force -> <Anonymous> -> roxy_meta_load
Execution halted
```

so unfortunately I think it just has to be an env rather than a list itself.

kevinushey:
```
=> Rcpp::compileAttributes()

* Updated R/RcppExports.R

==> devtools::check()

Updating roxygen2 documentation
Updating roxygen version in /Users/kevinushey/r/pkg/roxygen2/DESCRIPTION
Writing NAMESPACE
Loading roxygen2

 *** caught segfault ***
address 0xffffffff012bb38b, cause 'memory not mapped'

Traceback:
 1: wrapString(x, width = as.integer(width))
 2: wrap_string.default(object_usage(attr(block, "object")), width = 75L)
 3: wrap_string(object_usage(attr(block, "object")), width = 75L)
 4: topic_add_usage(rd, block)
 5: block_to_rd(block, base_path, env, global_options)
 6: roclet_process.roclet_rd(X[[i]], ...)
 7: FUN(X[[i]], ...)
 8: lapply(roclets, roclet_process, blocks = blocks, env = env, base_path = base_path,     global_options = options)
 9: roxygen2::roxygenise(pkg$path, roclets)
10: force(code)
11: withr::with_envvar(r_env_vars(), roxygen2::roxygenise(pkg$path,     roclets))
12: document(pkg)
13: devtools::check(, check_dir = dirname(getwd()))
An irrecoverable exception occurred. R is aborting now ...
```

😭

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # roxygen2 (development version)
 
+* A system for providing custom roxygen2 metadata is now available.
+  If the file "man-roxygen/roxygen-meta.R" is available, it will be
+  evaluated. Its evaluation should produce an R list, mapping roxygen
+  option names to values. Currently, only 'family.prefix' is recognized:
```
hadley:
```suggestion
  option names to values. Currently, only 'rd_family_title' is recognized:
```

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # roxygen2 (development version)
 
+* A system for providing custom roxygen2 metadata is now available.
+  If the file "man-roxygen/roxygen-meta.R" is available, it will be
```
hadley:
```suggestion
  If the file "man/roxygen/meta.R" is available, it will be
```

R/meta.R
```diff
@@ -0,0 +1,52 @@
...
+  pattern <- "^meta[.][Rr]"
+  meta_files <- list.files(meta_dir, pattern = pattern, full.names = TRUE)
+  if (length(meta_files) == 0)
+    return(FALSE)
+
+  if (length(meta_files) > 1) {
+    rlang::abort("Multiple 'man/roxygen/meta.R' files found")
+  }
+
+  parsed <- tryCatch(parse(meta_files), error = identity)
```
hadley: Oh just noticed that you're checking for the result of `tryCatch()`; I think that's generally bad practice and instead you should do:

```R
parsed <- tryCatch(
  parse(meta_files),
  error = function(cnd) {
     message <- "Could not parse 'man/roxygen/meta.R'"
     rlang::abort(message, parent = cnd)
  }
}

File: R/meta.R [hunk 16]
```diff
@@ -0,0 +1,52 @@
...
+  if (inherits(parsed, "error")) {
+    message <- "Could not parse 'man/roxygen/meta.R'"
+    rlang::abort(message, parent = parsed)
+  }
+
+  # TODO: appropriate evaluation environment?
```
hadley: `baseenv()` might be more appropriate? So that you have to qualify any package functions?

R/rd-family.R
hadley: I'm slightly worried that if it's not a list (e.g. it's a function) you'll get an error message that will be extremely difficult to diagnose. I'm not sure if it's worth handling that rare case.

tests/testthat/test-meta.R
```diff
@@ -0,0 +1,25 @@
+context("Meta")
+
+test_that("evaluation of 'man/roxygen/meta.R' must return a list", {
+
+  tmpdir <- tempfile("roxygen-meta-")
+  on.exit(unlink(tmpdir), add = TRUE)
```
hadley: Also `roxy_meta_set(list())` on exit?

R/meta.R
kevinushey: One thought I had: should `meta.R` be able to have access to functions used in your own package (e.g. your helper functions)?

hadley: My gut feeling is no — these should be fixed options. (And we can always relax in the future; it'd be hard to rollback to only evaluating in base env if we allow package env now)

hadley:
Thanks @kevinushey!

PR #886, opened 2019-07-24
omit already documented params from inheritDotParams, closes #885

### LINKED ISSUE #885: inheritDotParams includes duplicate param definitions if param already documented
In the (I think) common case where one wants to make a wrapper function (say `wrapper`) around an existing function (`original`) where some (but not all) parameters from `original` are set to particular values in `wrapper`, I often see (and am currently using) a pattern of overriding just those parameters and forwarding everything else with `...`. Like this:

```r
#' Wrapper around original
#'
#' @inherit original
#' @inheritDotParams original
#' @param y some more specific description
#' @export
wrapper <- function(x = "some_value", y = "some other value", ...) {
  original(x = x, y = y, ...)
}

#' Original function
#'
#' @param x x description
#' @param y y description
#' @param z z description
#' @export
original <- function(x, y, z, ...) {}
```

I think this is a good case for `@inherit` + `@inheritParams` simultaneously, and is related to but separate from #857 --- I'm currently encountering both problems in writing function wrappers. The issue is that @inheritDotParams includes parameters in its listing under `...` even if they are already defined in the function (either because of a `@param` usage or an `@inherit` usage). Thus I get this as output:

```
\arguments{
\item{x}{x description}

\item{y}{some more specific description}

\item{...}{Arguments passed on to \code{original}
\describe{
  \item{x}{x description}
  \item{y}{y description}
  \item{z}{z description}
}}
}
```

But would expect the output below instead, since `x` nor `y` are already defined by the function and so cannot be passed through `...`:

```
\arguments{
\item{x}{x description}

\item{y}{some more specific description}

\item{...}{Arguments passed on to \code{original}
\describe{
  \item{z}{z description}
}}
}
```

I know that I can get this output by adding `-x -y` to the `@inheritDotParams` spec, but it seems like this could be detected automatically without any false positives, and it would simplify documentation maintenance quite a bit in cases where there is more than just a parameter or two (I am currently working on some geoms that have an exclusion list of ~8 params, all of which I think it could exclude automatically).

mjskay:
This PR is a fix for #885. It changes `@inheritDotParams` slightly so that it only lists parameters under `...` that aren't already documented. 

Since parameter names that are listed as arguments to the function (and therefore should already be documented) can't get passed to `...` anyway (right?), this seems like a safe change that will not have unexpected side effects, but which can also save some maintenance cost for users when wrapping functions and passing arguments using `...`.


hadley:
Thanks!

Issue #891, opened 2019-07-29
External roclets


mikldk:
As noted in #865 it is possible to build a package with a custom roclet. I have done that in [mikldk/roxytest](https://github.com/mikldk/roxytest). Thanks for making that possible.

During the process I have experienced some issues that I wanted your comments on if possible. The use-case is a package using the above mentioned [`roxytest`](https://github.com/mikldk/roxytest) by adding the following lines to tje package's `DESCRIPTION` file:
```
Imports: roxygen2, testthat, roxytest
Roxygen: list(roclets = c("namespace", "rd", "testthat_roclet"))
```

Then the custom roclet can be used by running the following:
```r
library(roxytest)
roxygen2::roxygenise()
```
 
There are a few things that I wished were more consistent in the above and a comment on testing:

1. In `DESCRIPTION`, `Roxygen: list(roclets = c("namespace", "rd", "testthat_roclet"))` must be added. It would be more consistent to omit `_roclet` of `testthat_roclet.
2. It would be convenient to avoid the need to call `library(roxytest)` first. Maybe by specifying `roxytest::testthat` as roclet in `DESCRIPTION` by suitably changing `roxygen2::roclet_find()`.
3. Better test exposure: `roxygen2` uses `roxygen2::roc_proc_text()` that only takes one roclet. Some custom roclets introducing new tags (e.g. `@tests` by [`roxytest`](https://github.com/mikldk/roxytest)) require e.g. the rd roclet, too (to get the function name).

Do you have any comments on the above? I could try to make one or more pull requests to address the issues, but I thought it made sense to ask what you thought before diving into it.

mikldk:
1. Can be addressed by PR #892.
2. Can be addressed by myself by adding `roxytest` to `Depends` instead of `Imports`. Then `roxygen2::roxygenise()` works after `devtools::load_all(".")`, but I am not sure if this is a good way to do it. I cannot come up with others than this and letting `roxygen2` loading the namespace, but that's not a good idea in my opinion.
3. (Working on it...)

mstr3336:
Hi! 
This was closed. Did this get resolved?

mstr3336:
> Recently, you can solve 1-2 more elegantly with
> 
> ```
> Roxygen: list(roclets = c("namespace", "rd", "roxytest::test_roclet"))
> ```

Awesome! This is great, as I was contemplating writing an entire new package, just to wrap up my custom roclets functionality without cluttering the namespace, but now it won't be necessary :)

Issue #889, opened 2019-07-25
Look for options in man/roxygen/meta.R


hadley:
As well as in DESCRIPTION `Roxygen` field

PR #892, opened 2019-07-30
Proposoals for issue 891

mikldk:
These are proposals for addressing the comments described in #891.

### DIFF
--- a/R/roclet.R
+++ b/R/roclet.R
@@ -94,6 +94,12 @@ roclet_find <- function(x) {
   env$namespace <- namespace_roclet
   env$vignette <- vignette_roclet
 
+  # Making it possible to specify external roclets
+  # without the _roclet postfix.
+  if (!grepl(pattern = "_roclet$", x)) {
+    x <- paste0(x, "_roclet")
+  }
+
   expr <- parse(text = x)
   res <- eval(expr, env)
 
--- a/R/roclet.R
+++ b/R/roclet.R
@@ -120,15 +120,19 @@ is.roclet <- function(x) inherits(x, "roclet")
 #'
 #' Useful for testing.
 #'
-#' @param roclet Name of roclet to use for processing.
+#' @param roclets Roclets to use for processing.
 #' @param input Source string
 #' @param registry Named list of tag parsers
 #' @param global_options List of global options
+#'
+#' @return Note that if only one roclet provided,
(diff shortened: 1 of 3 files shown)


R/roclet.R
```diff
@@ -114,23 +120,48 @@ is.roclet <- function(x) inherits(x, "roclet")
...
 #' @param global_options List of global options
+#'
+#' @return Note that if only one roclet provided,
+#' the result will be simplified to its result.
+#' Otherwise, a list of results are returned.
+#'
 #' @export
 #' @keywords internal
-roc_proc_text <- function(roclet, input, registry = default_tags(),
+roc_proc_text <- function(roclets, input, registry = default_tags(),
```
hadley: I generally don't like this sot of interface because it makes it difficult to predict the output type of a function.

R/roclet.R
```diff
@@ -94,6 +94,12 @@ roclet_find <- function(x) {
   env$namespace <- namespace_roclet
   env$vignette <- vignette_roclet
 
+  # Making it possible to specify external roclets
+  # without the _roclet postfix.
+  if (!grepl(pattern = "_roclet$", x)) {
```
hadley: I'd prefer to encourage people to use the full names. When #889 is implementation, we can remove the special casing for these existing roclets.

R/roclet.R
mikldk: Then what about a new function called something like `roc_proc_text_lst()`?

R/roclet.R
mikldk: All good. Thanks for the input.

R/roclet.R
hadley: Why is it needed?

mikldk: So that roclets working together can be tested. For example to exploit tags processed by e.g. the `rd` roclet in other roclets, e.g. in my experimental [`roxytest::testthat_roclet`](https://github.com/mikldk/roxytest). In this I consider enabling the inclusion of `@example`(s) in the `testthat` skeleton (e.g. with `expect_error({ example }, NA)`) for the benefit of test coverage and maybe to avoid replication: a package developer can then write an example instead of a test, which perhaps results in better documentation.

But I am in no way intransigent; it is a suggestion that I wanted your feedback on. I fully acknowledge that the idea may be bad :-).

PR #894, opened 2019-08-08
Update links to point to r-lib/roxygen2

nealrichardson:


hadley:
Thanks @nealrichardson!

Issue #799, opened 2018-10-03
Rmd mode


hadley:
One of the challenges of writing longer documentation with Rd is that you can not mingle prose and code, as all code must live in the examples block. I suggest we create an new `@Rmd` tag which would treat everything afterwards like a Rmd document, including embedding the results of ```` ```{r} ```` blocks and converting `##` and `###` to sections and subsections.  You would still use existing roxygen tags for setting the title and parameters, but `@rmd` would provide an alternative for all other sections (including description, see also, return, and examples).

```
#' Combine vectors 
#' 
#' @param ... Vectors to coerce.
#' @rmd
#' ## Basic operation
#' 
#' `vec_c()` combines many small vectors into one long vector:
#' 
#' ```{r}
#' vec_c("a", "b", "c")
#' vec_c(1, 2, 3)
#' ```
#' 
#' If the inputs are of different types, `vec_c()` will coerce to a
#' common type:
#' 
#' ```{r}
#' vec_c(FALSE, 1, 1.5)
#' ```
#'
#' But note that not all pairs of vectors have a common type, and 
#' `vec_c()` will throw an error if a common type can not be found:
#' 
#' ```{r, error = TRUE}
#' vec_c(factor("x"), 1:10)
#' ```
#' 
#' ## Prototype
#' 
#' You can optionally specify a protoype, with the `.ptype` argument.
#' When this is specified, `vec_c()` no longer tries to find the common
#' type, but instead coerces all inputs to the supply type.
#' 
#' ```{r}
#' vec_c(1L, .ptype = double())
#' ```
```

Open questions:

* How to handle plot heavy packages like ggplot2, which has ~800 plots in examples? Should we render small thumbnails by default, and have some option to suppress all together?

* When should `document()` re-run all the inline code? Every time? Only optionally? Just before release? As part of some new function that tests and documents a single file?

hadley:
Or is it simply better to use vignettes for this task? Or with R-core to implement this in Rd?

mine-cetinkaya-rundel:
Using vignettes for this could result in many shorter vignettes, is that desirable? The important pieces of information about specifics of a package may not be easily discoverable if there are many vignettes. Could the docs link to specific sections of a longer vignette? (This might be possible now, it's not something I've tried.)

I think the idea behind this issue was to be able to move some of the most used pieces of information (code examples) further up in the docs without losing the technical details that also need to live there.

PR #902, opened 2019-09-10
Support including (R)md files into a manual page

gaborcsardi:
This facilitates sharing text between README/vignettes and manual pages.

TODO:
- [ ] Handle sections/headings properly.
- [ ] Add docs.
- [ ] Add more tests.


.Rbuildignore
```diff
@@ -14,3 +14,5 @@
 ^\.github$
 ^pkgdown$
 ^\.covrignore$
+^\.Rprofile$
```
hadley: Did you mean to include these?

R/rd-include-rmd.R
```diff
@@ -0,0 +1,35 @@
...
+block_include_rmd <- function(tag, block, env) {
+  stopifnot(is.character(tag), length(tag) == 1, !is.na(tag))
+  md_path <- tempfile(fileext = ".md")
+  on.exit(unlink(md_path, recursive = TRUE), add = TRUE)
+  if (!requireNamespace("rmarkdown", quietly = TRUE)) {
+    stop("@includeRmd requires the rmarkdown package")
+  }
+  rmd_path <- rmd_process_links(tag)
+  on.exit(unlink(rmd_path, recursive = TRUE), add = TRUE)
+  rmarkdown::render(rmd_path, output_format = rmarkdown::github_document(),
```
hadley: Can you please use tidyverse style indenting here? http://style.tidyverse.org/syntax.html#indenting

.Rbuildignore
gaborcsardi: Yep, if you don't mind. It makes my life easier and these should not be in the R package anyway, so it should be harmless for others.

R/rd-include-rmd.R
gaborcsardi: Ooops, sure.

gaborcsardi:
> that the Rmd is copied to a temporary directory (so you can't rely on local files), and that it's not cached (so you'll need to be careful to make it fast).

Yeah, let me think about this, it would be great to have caching by default. Unfortunately we need to make a copy to make the links to package objects work.

gaborcsardi:
As for the sectioning, I think it would be reasonable to do this:

- if there are no headings in the Rmd, then put the whole generated Rd into `\details{}`.
- if there are subsections in the Rmd (i.e. `##` and below), then create subsections inside `\details{}`.
- level 1 headings, i.e. `#`, go in their own `\section{}`s.
- if there is a level 1 heading, but not at the beginning of the Rmd file, then the beginning of the file goes to `\details{}`, up to the first `#`, and then we create new sections. 

I think this is pretty intuitive, and most of often this is what you want.

gaborcsardi:
@hadley What should be the working directory? The package dir? The directory of the R file being processed? 

Right now `document()` and `roxygenize()` do not change the working directory AFAICT. rmarkdown/knitr change it to the directory of the Rmd file.

We could change the working directory or try calculate the file name relative to the package directory. The former is simpler.

hadley:
I think we should consistently use the package directory as the working directory

gaborcsardi:
> I think we should consistently use the package directory as the working directory

OK, for now I'll switch for `@includeRmd` and then we can think about switching for `document()` and/or `roxygenize()`.

gaborcsardi:
Btw. `@includeRmd` has proper support for sections now. We could have the same for "regular" roxygen2 markdown, we just need to turn it now, basically. Then we could use `#` etc. for (sub)sectioning.

Re #799, at this point it is super easy to add an `@rmd` tag as well, basically the same as `@includeRmd()` but takes the text from the block, instead of the external file. It would need some conventions where the actual Rmd files are created to support caching, whether they would share an environment, etc. So I will not do this now, but I think we could keep #799 open and have another PR for it later.

gaborcsardi:
Actually, re paths, since rmarkdown/knitr defaults to set the wd to the directory of the Rmd, I think we should do the same, for the chunks of the Rmd. We would still switch to the package directory for running `rmarkdown::render()`, so the paths you specify in `@includeRmd` would be relative to that. Is this OK?

R/markdown-link.R
```diff
@@ -53,21 +52,21 @@ add_linkrefs_to_md <- function(text) {
...
 
-  paste0(
-    text,
-    "\n\n",
-    paste(ref_text, collapse = "\n"),
-    "\n"
-  )
+add_linkrefs_to_md <- function(text) {
+  ref_text <- get_md_linkrefs(text)
+  if (! length(ref_text)) return(text)
```
hadley: I suggest having `add_linkrefs_to_md()` return a character vector, since that will make testing a bit easier.  (Currently your check here is inconsistent with the early return above.)

R/markdown.R
```diff
@@ -128,3 +141,31 @@ mdxml_image = function(xml) {
   title <- xml_attr(xml, "title")
   paste0("\\figure{", dest, "}{", title, "}")
 }
+
+mdxml_heading <- function(xml, state) {
+  if (state$tag$tag != "@includeRmd") {
+    return(mdxml_unsupported(xml, state$tag, "markdown headings"))
```
hadley: Lets get this PR merged and then add heading support everywhere.

R/rd-include-rmd.R
```diff
@@ -0,0 +1,61 @@
+
+block_include_rmd <- function(tag, block, base_path) {
+  rmd <- tag$val
+  stopifnot(is.character(rmd), length(rmd) == 1, !is.na(rmd))
+  if (!requireNamespace("rmarkdown", quietly = TRUE)) {
```
hadley: `is_installed()` might be a little cleaner since we're using rlang already

R/rd-include-rmd.R
```diff
@@ -0,0 +1,61 @@
...
+    stop("@includeRmd requires the rmarkdown package")
```
hadley:
```suggestion
    abort("@includeRmd requires the rmarkdown package")
```

(For consistency this should probably be with a warning that gives the tag location, but this is likely to be a rare error so it's not worth it)

R/rd-include-rmd.R
```diff
@@ -0,0 +1,61 @@
...
+  rmd <- normalizePath(rmd)
+
+  cache_path <- paste0(sub("\\.Rmd$", "", rmd), "_cache/")
+  fig_path <- file.path(dirname(rmd), "figure/")
+  linkrefs <- rmd_linkrefs_from_file(rmd)
+  opts <- c(
+    root.dir = dirname(rmd),
+    cache.path = cache_path,
+    fig.path = fig_path,
+    child = rmd)
```
hadley:
```suggestion
    child = rmd
  )
```

R/rd.R
```diff
@@ -404,6 +406,22 @@ topic_add_eval_rd <- function(topic, block, env) {
   }
 }
 
+topic_add_include_rmd <- function(topic, block, base_path) {
+  rmds <- block_tags(block, "includeRmd")
+
+  for (rmd in rmds) {
+    tag <- roxy_tag("@includeRmd", rmd, attr(block, "filename"),
```
hadley: Needs tidyverse style

R/utils.R
```diff
@@ -173,3 +173,9 @@ pkg_env <- function() {
   env$.packageName <- "roxygen2"
   env
 }
+
+uuid <- function(nchar = 8) {
+  paste(
+    sample(c(letters, LETTERS, 0:9), nchar, replace = TRUE),
+    collapse = "")
```
hadley: And here

gaborcsardi:
> I was thinking mostly about the path to the .Rmd

Yes, that's now relative to the package.

I'll update and add the missing stuff in a minute....

R/markdown-link.R
gaborcsardi: Perhaps you mean `get_md_linkrefs()`. (`add_linkrefs_to_md()` has the text as a string as input already.) Yeah that makes sense, it was like that in a previous version, hence the incorrect early return.... 

Done in b4eb16de0082ba003e35e57f2cf6894b7dd422d9

R/markdown.R
gaborcsardi: OK.

R/rd-include-rmd.R
gaborcsardi: OK. I also moved this check a bit upstream.

R/rd-include-rmd.R
gaborcsardi: OK, made it into a warning.

hadley:
Can you please merge/rebase fix the failure?

gaborcsardi:
Rebased, but I think it'll still fail, because there is no pandoc.

PR #903, opened 2019-09-10
Conditionally generate \verb{} instead of \code{}

hadley:
`\code{}` is supposed to be used with [R code](https://cran.rstudio.com/doc/manuals/r-devel/R-exts.html#Marking-text):

> `\code{text}`
> Indicate text that is a literal example of a piece of an R program, e.g., a fragment of R code or the name of an R object. Text is entered in R-like syntax, and displayed using typewriter font where possible. Macros \var and \link are interpreted within text.

Previously we were using it to translate every use of backticks, which causes problems if you're attempting to use  `` `/` ``, `` `{` `` and `` `}` ``. This patch uses `\code{}` the contents of backticks parses, and otherwise falls backs to `\verb{}` (with more escaping).

### DIFF
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -38,14 +38,15 @@ mdxml_node_to_rd <- function(xml, tag) {
     document = ,
     unknown = mdxml_children_to_rd(xml, tag),
 
-    code_block = paste0("\\preformatted{", gsub("%", "\\\\%", xml_text(xml)), "}"),
     paragraph = paste0("\n\n", mdxml_children_to_rd(xml, tag)),
     text = xml_text(xml),
-    code = paste0("\\code{", gsub("%", "\\\\%", xml_text(xml)), "}"),
     emph = paste0("\\emph{", mdxml_children_to_rd(xml, tag), "}"),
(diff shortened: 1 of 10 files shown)


R/markdown.R
```diff
@@ -70,6 +71,32 @@ mdxml_unsupported <- function(xml, tag, feature) {
...
+  }, error = function(x) FALSE)
+}
+
+escape_verb <- function(x) {
+  x <- gsub("\\", "\\\\", x, fixed = TRUE)
+  x <- gsub("%", "\\%", x, fixed = TRUE)
+  x <- gsub("{", "\\{", x, fixed = TRUE)
+  x <- gsub("}", "\\}", x, fixed = TRUE)
+  x
+}
```
gaborcsardi: This seems to produce the correct output, almost always, which is amazing! The only glitch I found is that `\%` is shown as `\\%` in the man page, and the same with ```` ``` ````. But maybe the PR #904 will fix this?

hadley: Yeah, that's the job of #904.

PR #904, opened 2019-09-10
Escape Rd comments in markdown

### LINKED ISSUE #879: Parsing for % characters
Can roxygen include a warning when the resulting .Rd file is mis-formatted as a result of un-escaped `%` characters? (which end up as comments in the .Rd format)

This sometimes produces a warning for mismatched quotes or braces (e.g. #530, #742, #786), but this doesn't always seem to be caught until R CMD CHECK.

reprex below:

``` r
library(roxygen2)

lines <- c("#' compute a percentage",
           "#' @param x the % value to use",
           "#' @export",
           "f <- function(x, y) {x / 100 * y}")

roc_proc_text(rd_roclet(), lines)
#> $f.Rd
#> % Please edit documentation in RtmpcTEw4i/file6ca12af04873
#> \name{f}
#> \alias{f}
#> \title{compute a percentage}
#> \usage{
#> f(x, y)
#> }
#> \arguments{
#>   \item{x}{the % value to use}
#> }
#> \description{
#> compute a percentage
#> }
```

<sup>Created on 2019-07-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
I think this is a worthwhile change — in order to be backward compatible `\%` will still generate `\%` so if you want to generate `\\%` you'll need to put it in a code block (or jump through considerable gymnastics). But I think this happens rarely in practice, while it's easy to forget that you need to escape `%`. If you do write `5%`, error messages are likely to be confusing because they'll usually come from invalid `.Rd` syntax (e.g. because a closing `}` is eaten by the comment).

Fixes #879

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # roxygen2 (development version)
 
+* Rd comments (`%`) are automatically escaped in markdown text. Existing
+  code that uses `\%` is automatically repaired; if you're generating
+  more exotic sequences like `\\%` you'll need to put them inside 
+  backticks (#879)
+
 * As well as storing roxygen options in the `Roxygen` field of the 
   `DESCRIPTION` you can now also store them in `man/roxygen/meta.R` (#889).
 
--- a/R/markdown-link.R
+++ b/R/markdown-link.R
@@ -137,7 +137,7 @@ parse_link <- function(destination, contents) {
     )
 
   } else {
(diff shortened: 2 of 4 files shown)


hadley:
Ok, whole new strategy, which is a little simpler than your idea, but still conveys the same message.

tbates:
Guessing that people easily miss this 1-time warning. I did.

Be nice to just always flag each line containing  `\%` so they can be weeded out as the coder has time

PR #907, opened 2019-09-13
Allow subsections in markdown

gaborcsardi:
This is a trivial change now. Level 1 heading support is coming in another PR.


gaborcsardi:
We should probably add something about this to the docs. Or we can also wait with the docs until level 1 headings are done.

hadley:
Thanks!

PR #908, opened 2019-09-13
Support level 1 headings in markdown

gaborcsardi:
They generate separate sections.

This PR is on top of #907.

I'll add docs later, once we make sure that this is what we want.


gaborcsardi:
Rebased.... it addition to the docs, I'll also add better tests for edge cases, like #909.

R/rd.R
```diff
@@ -41,16 +41,16 @@ roclet_tags.roclet_rd <- function(x) {
...
+    details = tag_markdown_with_sections,
     docType = tag_name,
     encoding = tag_value,
     evalRd = tag_code,
     example = tag_value,
     examples = tag_examples,
     family = tag_value,
     field = tag_name_description,
-    format = tag_markdown,
+    format = tag_markdown_with_sections,
```
hadley: I think we should only allow in `@description` and `@details` for simplicity

R/rd.R
```diff
@@ -243,7 +243,10 @@ topic_add_simple_tags <- function(topic, block) {
   tag_names <- names(block)[is_simple]
 
   for (i in seq_along(tag_values)) {
-    topic$add_simple_field(tag_names[[i]], tag_values[[i]])
+    topic$add_simple_field(tag_names[[i]], tag_values[[i]][1])
```
hadley:
```suggestion
    topic$add_simple_field(tag_names[[i]], tag_values[[i]][[1]])
```

R/tag.R
```diff
@@ -241,3 +241,21 @@ tag_markdown <- function(x) {
   x$val <- markdown_if_active(x$val, x)
   tag_value(x)
 }
+
+#' @export
+#' @rdname roxy_tag
+tag_markdown_with_sections <- function(x) {
+  x$val <- markdown_if_active(x$val, x, sections = TRUE)
+  if (length(x$val) == 1 && x$val[[1]] == "") {
+    roxy_tag_warning(x, "requires a value")
```
hadley: Can you use the same test as above? And it needs to return, so that further processing doesn't occur.

R/tag.R
```diff
@@ -241,3 +241,21 @@ tag_markdown <- function(x) {
...
+  }
+  for (i in seq_along(x$val)) {
+    if (!rdComplete(x$val[i])) {
```
hadley: Why doesn't this match `tag_markdown()`?

R/rd.R
gaborcsardi: OK.

R/rd.R
gaborcsardi: `tag_values[[i]]` is a character vector now, for tags with sections. This code is not the easiest to read, so we can make sections more explicit if you want. Then we'll need a `topic_add_tags_with_sections()` function I guess.

R/tag.R
gaborcsardi: Hmmm, which test? As in `tag_value()`? `x$val` can be a vector now, so it needs a slightly different test.

R/tag.R
gaborcsardi: Because `x$val` is a vector now, First element is the regular content, the rest are the sections. But again, we can make this more explicit by returning a list.

R/rd.R
gaborcsardi: done.

R/tag.R
hadley: I think it would be natural to match `tag_examples()`; i.e. check `x$val == ""` before parsing the markdown.

R/tag.R
hadley: But `tag_markdown()` doesn't perform this check at all. Why not?

R/tag.R
gaborcsardi: oh, ok, that makes sense.

R/tag.R
gaborcsardi: It does I think, it calls `tag_value()`, that does these checks.

hadley: Ah got it.

hadley:
I just pushed a little tweak. I'll merge once you update the docs.

gaborcsardi:
OK, added short docs in the Rd formatting vignette.

gaborcsardi:
@hadley also rebased, so it is ready, I think.

hadley:
Thanks!

PR #914, opened 2019-09-18
Expose three code loading strategies

### LINKED ISSUE #822: Error: must pass a package name, “DLLInfo” or “DllInfoReference” object
The recent v6.1.1 update causes the following error on Travis:

https://travis-ci.org/mlr-org/mlr/jobs/451819776#L2685

```r
Running script: step_run_code(devtools::document())
Updating mlr documentation
Updating roxygen version in /home/travis/build/mlr-org/mlr/DESCRIPTION
Writing NAMESPACE
Loading mlr
Loading required package: ParamHelpers
Error: must pass a package name, “DLLInfo” or “DllInfoReference” object
```

The error does not appear with v6.1.0. Also it just appeared today after v6.1.1 was released on CRAN.

Using `devtools::document(".")` works in some instances. I think this could be related to the change made in #704?

hadley:
Fixes #822


R/load.R
```diff
@@ -0,0 +1,92 @@
...
+# Helpers -----------------------------------------------------------------
+
+find_load_strategy <- function(x, options) {
+  if (is.function(x)) {
+    return(x)
+  }
+
+  if (is.null(x)) {
+    x <- x %||% options$load
+  }
```
gaborcsardi: Don't need the `if` here, I believe.

R/roxygenize.R
```diff
@@ -8,25 +8,26 @@
...
-#'   This defaults to `NULL`, which will use the `roclets` fields in
-#'   the list provided in the `Roxygen` DESCRIPTION field. If none are
-#'   specified, defaults to `c("collate", "namespace", "rd")`.
+#'   The default, `NULL`, uses the roxygen `roclets` option,
+#'   which defaults to `c("collate", "namespace", "rd")`.
 #' @param load_code A function used to load all the R code in the package
-#'   directory. It is called with the path to the package, and it should return
-#'   an environment containing all the sourced code.
+#'   directory. The default, `NULL`, uses the strategy defined by
+#'   the `code` roxygen option, which defaults to [load_pkgload()].
```
gaborcsardi: `code` roxygen option -> `load` roxygen option

R/load.R
```diff
@@ -0,0 +1,92 @@
...
+#'
+#' You can change the default strategy for your function with roxygen2 `load`
+#' option. Override the default off `pkgload` to use the `source` or
+#' `installed` strategies:
+#'
+#' ```
+#' RoxygenNote: list(load = "source")
+#' ```
+#' @name load
+#' @param path Path to source package
```
gaborcsardi: Can this be a subdirectory, or it has to be the root?

DESCRIPTION
```diff
@@ -58,5 +58,5 @@ LinkingTo:
 VignetteBuilder: 
     knitr
 Encoding: UTF-8
-Roxygen: list(markdown = TRUE)
+Roxygen: list(markdown = TRUE, load = "installed")
```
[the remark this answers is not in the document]
hadley: Yes, because roxygen2 is one of those packages that crashes if you reload it 😬

R/load.R
hadley: Has to be root

R/load.R
```diff
@@ -0,0 +1,92 @@
...
+NULL
+
+#' @rdname load
+#' @export
+load_pkgload <- function(path) {
+  pkgload::load_all(path, helpers = FALSE, attach_testthat = FALSE)$env
```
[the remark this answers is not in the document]
hadley: Yeah, because that caused problems. Re-reading the history of the problem, I first attempted to not compile because DE complained that it was compiling with different build flags. But that change meant that it didn't compile when it was actually needed, which caused more problems that it solved. Now, if the default pkgload strategy doesn't work, you can roll back to the pure source strategy.

R/load.R
hadley: Ooops

R/load.R
hadley: `devtools::document()` will find the package root though

gaborcsardi:
For `load_installed`, do you think there is a way to detect if the installed package is outdated? Maybe we could use the `Built` time stamp from the installed `DESCRIPTION`, and the latest time stamp of the files in `R`?

I guess "installed" should not be common, so it might not be worth it.

hadley:
@gaborcsardi yeah, I think if you use the installed strategy you're on your own.

jimhester:
This looks good, thanks for working on it!

PR #919, opened 2019-09-20
Markdown improvements

gaborcsardi:
* Allow HTML blocks and inline HTML in Rmd files included via `@includeRmd`. They are converted to `\if{html}{\out{...}}`.
* Allow markup in markdown headings. E.g. 
    ```
    # This is a **strong** heading
    ```
* Allow markup in markdown links. E.g.
    ```
    This is a [`link to an R object`][object].
    This is an [`external link`](https://external.link)
    ```
* Keep language info from fences.
    ````
    ```r
    ...
    ```
    ````
  will generate
    ```
    <div class="r"><pre>...</pre></div>
    ```
  in HTML.

Maybe we could mention some of these in the NEWS?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -74,6 +74,10 @@
 * Rd comments (`%`) are automatically escaped. You will need to replace any
   existing uses of `\%` with `%` (#879).
 
+* Markdown links can now contain formatting. E.g. in
+  `See [*Formatting*][formatting]` the link text will be emphasised.
+  Similarly in external links: `See [*this web page*](https://...)` (#919).
+
 * Use of unsupported markdown features (blockquotes, headings, inline HTML, 
   and horizontal rules) now gnerate more informative error messages (#804).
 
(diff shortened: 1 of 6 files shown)


gaborcsardi:
@hadley hmmm, I'll take a look at the test failures...

gaborcsardi:
@hadley OK, fixed the tests.

gaborcsardi:
@hadley Rebased this, it should fix the test failures.

hadley:
Would you mind adding a couple of news bullets? I think the two most important changes are allowing inline html and formatting in links.

gaborcsardi:
Well, the HTML is for `@includeRmd` only I believe, which is also new, so it does not need NEWS. 

I did not turn it on for non-Rmd markdown, because I was worried that things like "a<b and c>b" will be detected as HTML markup, and that will affect the existing documentation.

gaborcsardi:
I'll add a bullet for the markup in links.

gaborcsardi:
@hadley OK, added NEWS bullet for markup in link text.

gaborcsardi:
@hadley Let me rebase again, with the R6 in master.

gaborcsardi:
@hadley OK, rebased, tests seem to be still fine.

PR #926, opened 2019-10-04
Omit empty R6 docs secs w/o documented fields

gaborcsardi:
I.e. omit the section not only if the class has no
public fields, but also if all fields are undocumented.


hadley:
Thanks!

PR #935, opened 2019-10-05
R6 inherited methods in <details>

### LINKED ISSUE #932: Request: Option to disable full list of R6 methods of superclass
I've just converted [minihtml](https://github.com/coolbutuseless/minihtml) over to latest roxygen2 from github to test out the R6 support and it's working pretty well.
 
In my particular case, displaying the full list of inherited methods (over 100 methods) is a bit too much, and all the "noise" detracts from this particular class's documentation.   Would it be possible/feasible to just have a single link to the superclass documentation instead of the exhaustive list of all methods?

Extreme Example: Here is a link to the online docs for `minihtml::HTMLDocument`, where the ~100 methods on the superclass really obscure the documentation of this class: 
https://coolbutuseless.github.io/package/minihtml/reference/HTMLDocument.html

---
Sidenote: Whether or not it's a good idea to generate a method for each of the ~100 HTML tags for `HTMLElement` class is a different conversation :)

gaborcsardi:
The tag is open by default if there are no more than
5 inherited methods.

Closes #932.

### DIFF
--- a/R/rd-r6.R
+++ b/R/rd-r6.R
@@ -250,8 +250,14 @@ r6_inherited_method_list <- function(block, r6data) {
 
   super_meth <- super_meth[rev(seq_len(nrow(super_meth))), ]
 
-  c("\\if{html}{\\subsection{Inherited methods}{",
-    if (nrow(super_meth) <= 5) "\\out{<details open>}" else "\\out{<details>}",
+  details <- paste0(
+    "<details ",
+    if (nrow(super_meth) <= 5) "open ",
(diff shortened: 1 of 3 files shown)


gaborcsardi:
@hadley Thanks! I'll improve the styling a bit.

gaborcsardi:
Looks like this now: 

Screen Shot 2019-10-05 at 2 19 34 PM

Could be better, but AFAICT there is no easy way to apply the same style to the details tag as for a `<h4>`.

Issue #933, opened 2019-10-05
R6 links to documentation for inherited methods not compatible with 'pkgdown'


coolbutuseless:
When the documentation pages are generated in 'pkgdown' the links to the superclass methods in the "Inherited methods" section don't work.

For example: Try any of the 'inherited methods' links on this page:
https://coolbutuseless.github.io/package/minihtml/reference/HTMLDocument.html

The URL for the documentation is https://coolbutuseless.github.io/package/minihtml/html/HTMLElement.html#method-as_character_inner when it probably should be https://coolbutuseless.github.io/package/minihtml/reference/HTMLElement.html#method-as_character_inner

I think it has to do with the use of the relative path link for inherted methods i.e. `\item \href{../../minihtml/html/HTMLElement.html#method-as_character_inner}{\code{minihtml::HTMLElement$as_character_inner()}}`

Not sure if this is really a roxygen2 issue, or if it's something pkgdown should be translating during the build process.

gaborcsardi:
Yeah, I think we need to work out what to do in this case. I am afraid that we need to make pkgdown smarter about these links. CC @hadley

hadley:
This is the flaw of hacking in the links. Maybe roxygen2 should add a class attribute to the link so that pkgdown knows to handle them specially?

gaborcsardi:
@hadley Yeah, that's what I was thinking. May it can add a class, plus a data attribute that contains the name of the package, class and method, and then pkgdown can handle it nicely.

hadley:
Sure, although it's also easy enough to parse that data from the url.

gaborcsardi:
With #936 merged roxygen2 now adds information that pkgdown can parse and then create the right links.

PR #936, opened 2019-10-05
Add HTML class and data to inter-package anchored links

gaborcsardi:
These currently happen when linking to inherited R6
methods.

Closes (the roxygen part of) #933.


gaborcsardi:
@hadley: Thanks!

PR #939, opened 2019-10-07
Add name of package being documented to roxy_meta

gaborcsardi:


R/options.R
```diff
@@ -30,18 +31,22 @@ load_options <- function(base_path = ".") {
...
 
   if (is.na(desc_opts)) {
-    list()
+    opts <- list()
   } else {
-    eval(parse(text = desc_opts), child_env(baseenv()))
+    opts <- eval(parse(text = desc_opts), child_env(baseenv()))
   }
+
+  list(opts = opts, package = desc_pkg)
```
hadley: This seems a bit too complicated. Could you just inject `package` directly into the returned list here?

gaborcsardi: OK, done.

gaborcsardi:
Thanks!

PR #943, opened 2019-10-15
Fix hashing files with Unicode paths, on Windows

gaborcsardi:
This fixes hashing files with Unicode paths, and thus `same_contents()` as well. Both `digest::digest()` and `tools::md5sum()` are broken in this case, the latter returns `NA`. We need to read in the contents and then hash that if this happens.

Note that this does not completely fix roxygen2 on Unicode paths, because base R has a lot of problems working with them in general. I have added a test case for this, it currently fails, so it is skipped.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -91,12 +91,26 @@ same_contents <- function(path, contents) {
 
   contents <- paste0(paste0(contents, collapse = "\n"), "\n")
 
-  text_hash <- digest::digest(contents, serialize = FALSE)
+  text_hash <- md5_string(contents)
 
-  path <- normalizePath(path, mustWork = TRUE)
-  file_hash <- digest::digest(file = path)
+  file_hash <- md5_file(path)
 
   identical(text_hash, file_hash)
+}
+
+md5_string <- function(text) {
+  digest::digest(text, serialize = FALSE, algo = "md5")
+}
+
+md5_file <- function(path) {
+  # tools::md5sum returns NA if it cannot open the file. This also happens
+  # for Unicode file names on Windows, so we try a bit harder.
+  md5 <- tools::md5sum(path)[[1]]
+  if (is.na(md5)) {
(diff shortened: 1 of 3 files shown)


gaborcsardi:
@hadley should we try to detect this situation up front, and suggest using `subst`?

hadley:
Is there any way we can munge the path so it will work when passed to a base function?

gaborcsardi:
> Is there any way we can munge the path so it will work when passed to a base function?

Not that I know of. I tried `shortPathName()`, it did not help. I think we can probably fix this by switching to fs for all file manipulation.

hadley:
We might as just wait for the new digest right? (Since @jimhester's patch got merged)

gaborcsardi:
I think so. So the alternative is a version requirement for digest, but we could still keep the test cases.

gaborcsardi:
I can update this PR in couple of days, when the digest windows binary is built.

gaborcsardi:
The thing is, the new digest has a bug, and it fails on non-unicode non-Windows platforms, so it is not ideal to switch to that now. See the error on Debian-clang: https://cran.rstudio.com/web/checks/check_results_digest.html

```
     Running test_encoding.R............... 0 tests Error in file(f, open = "wb") :
     unable to translate '/tmp/Rtmp9L0NVP/file7c9d366409dd/tricky-<U+0151>' to native encoding
     Calls: <Anonymous> ... run_test_dir -> lapply -> FUN -> eval -> eval -> file
     Execution halted
```

gaborcsardi:
@jimhester sure, but the thing is, roxygen/base is still broken at several places for Unicode file names on Windows, so it is actually not urgent to fix this. We can aim to fix all of these issues for the release after the one coming out soon.

gaborcsardi:
I think we can close this, digest is fixed, so this is not needed. I still aim to fix all base R and roxygen issues with Unicode Windows paths mid-term.

PR #953, opened 2019-10-31
Add `r6` option to suppress special handling of R6 classes

### LINKED ISSUE #949: Add a global option that turns of the new R6 docs generation globally
To let people transition at their own pace.

gaborcsardi:
Closes #949.


gaborcsardi:
Thanks!

PR #960, opened 2019-11-06
Small typo in brackets

msberends:
One bracket should be removed.

One other thing though:
> Apart from `@keywords internal`, these tags are not very useful because most people find documentation using Google. 

I disagree here, especially after feedback from our users. Use case: we have a function that determines if bacteria are multi-resistant, called `mdro()`. The outcome could be MDR (multi-drug-resistant) or XDR (extensively drug-resistant). If a users puts in `?XDR` we would like them to get to the `mdro()` page that tells about it, so we added `@aliases MDR XDR`. Seems *very* useful to me and is very much faster than searching Google. So I think you're underestimating aliases here :)

### DIFF
--- a/vignettes/rd.Rmd
+++ b/vignettes/rd.Rmd
@@ -639,7 +639,7 @@ Three other tags make it easier for the user to find documentation within R's he
 
 *   Use `@keywords keyword1 keyword2 ...` to add standardised keywords. 
     Keywords are optional, but if present, must be taken from the 
-    predefined list found `file.path(R.home("doc"), "KEYWORDS"))`.
+    predefined list found `file.path(R.home("doc"), "KEYWORDS")`.
 
 Apart from `@keywords internal`, these tags are not very useful because most people find documentation using Google. `@keywords internal` is useful because it removes the function from the documentation index; it's useful for functions aimed primarily at other developers, not typical users of the package.
 


hadley:
What does `mdro()` return? A string?

How does a user know that `?XDR` will work? Why wouldn't they do `?mdro`?

msberends:
`mdro()` returns a factor with ordered levels: Negative < MDR < XDR. XDR is more severe than MDR. The definitions of MDR and XDR are from guidelines known by users. 

People wouldn't know *a priori* if `?XDR` would work, but they might look for a function that determines if a bug is an XDR, and use `?XDR` as input. It is a more common abbreviation in our field than MDRO (`mdro()`, *multi-drug-resistant organism*), but MDRO covers the whole diversity of resistant bugs, whereas XDR is only one outcome. So having `mdro()` as a function makes sense, and `XDR()` wouldn't since it is only one of the severities of resistance.

Stating "these tags are not very useful" is too short a turn, IMHO.

hadley:
I'd be happy to change to "rarely useful", but I don't think that's a terribly big difference.

msberends:
True :)
Maybe it’s my scientific brain that doesn’t like the statement. But you can of course leave it as it is, I understand the benefits are very small, if at all.

PR #982, opened 2019-11-21
Allow other section for @includeRmd

### LINKED ISSUE #970: Expand `@includeRmd` in the description section
From @gaborcsardi: We can't have nested roxygen tags currently, but the section to expand to could be a parameter.

gaborcsardi:
Closes #970.


lionel-:
Does this work with arbitrary sections? Or could it be made to?

gaborcsardi:
Could be anything, yes, roxygen will generate it at least, but it is not necessarily valid Rd. E.g. you can only have subsections within certain sections, etc.

lionel-:
With

```r
#' @section This is a section:
```

How do we expand in the section?

gaborcsardi:
Oh, that you cannot. Only the default Rd sections, not custom sections. In that case you can put the whole section in the Rmd as a workaround.

vignettes/rd.Rmd
```diff
@@ -586,7 +586,9 @@ Starting from roxygen2 7.0.0, you can use `@includeRmd path/to/file.Rmdname` to
 `@includeRmd` supports headings in the external Rmd. The rules are as follows:
 
 *   All text before the first level 1 heading (i.e. `#`), is added to the 
-    details section.
+    details section. If you prefer a different section, then write the
+    name of the section after the path in the `@includeRmd` tag, in all
+    lowercase. Example: `@includeRmd path description`.
```
hadley: I think you need to add a note that this only works with non-user defined sections (I don't know if we have a better name for that)

gaborcsardi: done.

lionel-:
Thanks for the quick fix!

Issue #990, opened 2019-11-25
Double backslashes in examples render differently with 7.0.1


wch:
I'm actually not sure if this is intentional or not. But in roxygen2 <= 7.0.0, given this input in the examples section:

```R
#' "\\alpha" \\
```

It would produce this output:

```
"\\\\alpha" \\\\
```

In 7.0.1, it produces this:

```
"\\alpha" \\
```

Reprex (note that there are extra backslashes in the input for escaping):

```R
library(roxygen2)
roc_proc_text(rd_roclet(), "
#' Example
#' @examples
#' \"\\\\alpha\" \\\\
f <- function() {}
")
#> $f.Rd
#> % Generated by roxygen2: do not edit by hand
#> % Please edit documentation in ./<text>
#> \name{f}
#> \alias{f}
#> \title{Example}
#> \usage{
#> f()
#> }
#> \description{
#> Example
#> }
#> \examples{
#> "\\alpha" \\
#> }
```

For a real-life example, the upgrade to 7.0.1 resulted in this change, which led to a R CMD check failure:

https://github.com/rstudio/shiny/commit/0cbe4bb3d4ed608539692c9e1b76f8884043a1c9#diff-fd9eb6354a4cc65e5016dd2dd015e7d4R20

If this is the intended behavior, then that's fine; we can change the .R files to match.

gaborcsardi:
Not intended. @hadley is aware of it, and I believe that there will be another patch version soon.

hadley:
Now I see 

``` r
library(roxygen2)

roc_proc_text(rd_roclet(), "
#' Example
#' @examples
#' \"\\\\alpha\" \\\\
f <- function() {}
")[[1]]$get_section("examples")
#> \examples{
#> "\\\\alpha" \\
#> }
```

<sup>Created on 2019-11-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #992, opened 2019-11-25
More escaping examples

hadley:


hadley:
Fixes #990

src/escapeExamples.cpp
```diff
@@ -0,0 +1,50 @@
...
+      } else if (*cur == '\'' || *cur == '"' || *cur == '`') {
+        in_string = *cur;
+      }
+    } else { // inside string/symbol
+      if (in_escape) {
+        in_escape = false;
+        if (*cur == 'l' || *cur == 'v') {
+          out += '\\';
+        } else if (*cur == '\\') {
+          out += "\\\\";
```
gaborcsardi: So this will add two backslashes, with `*cur` it is three, which does not seem right. Why does this work?

gaborcsardi: Never mind, got it finally. We wrote out the original escape already, and three more is four.

hadley: Yeah, I don't understand it either but empirically it works.

PR #993, opened 2019-11-25
Add @examplesIf tag

### LINKED ISSUE #962: Dynamic examples, take two
PoC implementation is here: https://github.com/gaborcsardi/dynex#readme

- Uses `\dontshow{}` tags to hide the check for the condition.
- No more `\Sexpr{}` tags, so don't need to install the package for `R CMD build`.
- Multiple `@dynex` tags work fine.

To include this in roxygen2, we would need to just copy over the code, probably rename the tag to `@examplesIf`, and add support for this in R6 classes.

gaborcsardi:
Quite simple. Let me know what you think.

Closes #962.


gaborcsardi:
@hadley Good as it is, well, except that it needs NEWS and docs, if we decide to go with this. FWIW I use it in several packages and seems to work well.

krlmlr:
If every line of code has its own `if()` block -- do we still need `withAutoprint()` and braces?

``` r
foo <- TRUE
if (foo) "foo"
#> [1] "foo"
if (foo) if (TRUE) "bar" else "baz"
#> [1] "bar"
if (foo) if (FALSE) "bar" else "baz"
#> [1] "baz"
```

<sup>Created on 2020-06-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

gaborcsardi:
@krlmlr hmmm, it is no obvious to me that you can do that. E.g. how about multi-line strings? Or raw strings? You could parse the expression, but this is starting to get much more complicated than `withAutoprint()`. I am not even sure what is wrong with `withAutoprint()`?

krlmlr:
Got you. I thought this pattern is applied to every statement. In this case `withAutoprint()` is just fine. I'll test-drive in my packages.

gaborcsardi:
@krlmlr Thanks!

hadley:
Can you add a news bullet? I'd like to merge this just so we can experiment a bit more with it (without advertising it heavily)

gaborcsardi:
(Closed by mistake.)

@hadley So, this version does not handle examples in R6 classes, which a bit special. It would need this commit to handle them: https://github.com/gaborcsardi/roxygenlabs/commit/fc249f5a931a7fbf403ac94a4424a8bdee5208e2

Should I add that?

hadley:
I think we should leave R6 for now

gaborcsardi:
@hadley Added NEWS bullet.

Issue #998, opened 2019-12-02
specify example in function body


bgctw:
proposed feature: new tag `@exampleFunction <functionName>`

where the body of referenced function <functionName> specifies the example code.
The approach combines several advantages:

-  specify example nearby the original code (instead of is own file with `@example`)
- use code formatting/ interactive execution and all the editor features (instead of commented
  code as with `@examples`)

This feature helps users transition to roxygen who use a similar feature a lot in their existing inlinedocs documentation.

Example:
```
#' Calculate mu and sigma of lognormal from summary statistics.
#' 
#' @exampleFunction example_getParmsLognormForModeAndUpper
getParmsLognormForModeAndUpper <- function(...){
}
example_getParmsLognormForModeAndUpper <- function(){
  # example 1: a distribution with mode 1 and upper bound 5
  (thetaEst <- getParmsLognormForModeAndUpper(1,5))
}
```

copying the lines from the body to the examples section:
```
  # example 1: a distribution with mode 1 and upper bound 5
  (thetaEst <- getParmsLognormForModeAndUpper(1,5))
```

gaborcsardi:
> specify example nearby the original code (instead of is own file with @example)

It is already near if you use `@examples`. It is typically right above the function, within the rest of the documentation.

> use code formatting/ interactive execution and all the editor features (instead of commented code as with @examples)

Most editors help you with this. E.g. in Emacs ESS and RStudio you can easily run the examples from the roxygen comments.

PR #999, opened 2019-12-02
implement @exampleFunction tag

bgctw:
tackles #998

There are still things that need improvment which need deeper roxygen2 expertise.

- Currently, the DESCRIPTION file is parsed many times in `roxy_tag_rd.roxy_tag_exampleFunction`, because I do not know how to access the name of the package that is currently documented from the this function.
- For avoiding duplication of writing, argument 'default' constructs the name of the function by prepending 'example_'  in front of the name of the documented function. Currently, it assume that the latter name is passed as `ls(env)[[1]]`, because I do not know how to access it in a better way.


hadley:
I think this is out of scope for roxygen2, but it would be a great feature for an extension package!

PR #1003, opened 2019-12-06
Convert embedded newlines as well

jimhester:
In some cases write_lines takes a vector of lines to write, in others the lines are already collapsed, so we have to rewrite the embedded newlines as well.

Fixes https://github.com/r-lib/roxygen2/issues/989

### DIFF
--- a/R/utils-io.R
+++ b/R/utils-io.R
@@ -12,6 +12,9 @@ write_lines <- function(text, path) {
     line_ending <- "\n"
   }
 
+  # we need to convert any embedded newlines as well
+  text <- gsub("\r?\n", line_ending, text)
+
   path <- file(path, open = "wb")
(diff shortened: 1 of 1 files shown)


jimhester:
@krlmlr if you want to fix `write_if_different()` I think that is the outstanding issue with this, I guess it would have to convert the newlines in one direction or the other before the comparison.

krlmlr:
Yeah, I wonder if we should compare the normalized output of `read_lines()` instead of doing `digest::digest()`.

JoFAM:
Had the same issue, installed roxygen2#1003, and can confirm the Rd files are no longer changed. I can also confirm the "Writing ...Rd" messages are still shown. 

Thank you for solving this.

PR #1009, opened 2019-12-13
Fix crash when @description or @details are NULL

### LINKED ISSUE #1008: roxygen2 reported error after upgrade from 6.1.1 to 7.0.2
Hi there 
    I have been using roxygen2 6.1.1 for all my R packages. Now after upgrade roxygen2 to version 7.0.2, I got the following error:

 Error in if (self$has_section(type) && !overwrite) { : missing value where TRUE/FALSE needed

   I reinstalled the old version 6.1.1 to keep updating the projects. 
Since I focus my attention on the maths and not in the documentation, perhaps I am missing some important detail.
 
You can check the project https://github.com/genomaths/MethylIT/tree/MethylIT_0.3.2_beta (roxygen2 6.1.1)

Would you please give me some tip on how to solve this problem?

Best,
Robersy

gaborcsardi:
Now they are ignored, except for `@description NULL` in
package level documentation, where it can be used to suppress the
auto-generated Description section

Closes #1008.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,11 @@
 * roxygen2 now keeps using Windows (CR LF) line endings for files that
   already have CR LF line endings, and uses LF for new files (#989).
 
+* roxygen2 now does not fail for `@description NULL` and `@details NULL`.
+  Instead, these tags are ignored, except for `@description NULL` in
+  package level documentation, where it can be used to suppress the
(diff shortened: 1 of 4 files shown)


gaborcsardi:
The CI failure on R-devel (macOS) is already there in master, so I'll merged this.

PR #1023, opened 2020-01-07
Improve error message for R6 in extract_r6_methods()

jameslamb:
Thanks for adding support for `R6`!  I've been looking forward to it for a while and I'm excited that it's here.

In this PR, I propose adding a bit more information to an error message I encountered the first time that I tried `roxygen2`  7.0.x with a package that has `R6` objects. It took me a while to go from the error message *"R6 class without source references."*, work back through the `roxygen2`  source code, and figure out that I needed to install my package with `--with-keep.source`. I think my proposed change would help others deal with this error a bit more easily.

### DIFF
--- a/R/object-r6.R
+++ b/R/object-r6.R
@@ -44,7 +44,16 @@ extract_r6_methods <- function(x) {
     x$public_methods[method_nms],
     function(m) {
       ref <- utils::getSrcref(m)
-      if (is.null(ref)) stop("R6 class without source references. Try re-installing with installation option '--with-keep.source'.")
+      if (is.null(ref)) {
+        name <- x$classname %||% NA
+        stop(
+          "R6 class ", if (!is.na(name)) paste0("(", name, ") "),
+          "without source references. ",
+          "If you use the `installed` load method in `DESCRIPTION`, then ",
+          "try re-installing the package with option '--with-keep.source'. ",
+          "E.g. `install.packages(..., INSTALL_OPTS = \"--with-keep.source\")`."
+        )
+      }
       utils::getSrcLocation(ref)
     }
   )
--- a/R/object-r6.R
+++ b/R/object-r6.R
@@ -44,7 +44,7 @@ extract_r6_methods <- function(x) {
     x$public_methods[method_nms],
     function(m) {
       ref <- utils::getSrcref(m)
-      if (is.null(ref)) stop("R6 class without source references")
+      if (is.null(ref)) stop("R6 class without source references. Try re-installing with installation option '--with-keep.source'.")
       utils::getSrcLocation(ref)
     }
   )


gaborcsardi:
Thanks! I agree we should improve this. Ideally we would change some other things as well.

1. I don't know how many people install packages from the command line. I would think most people use an IDE or `install.packages()`. On Windows, people probably don't use the shell much. So mentioning that option might be confusing.

2. roxygen does not use the installed package by default but it uses its own package code loading mechanism. Are you documenting the installed package? The reason for the current minimal package is that it is a safeguard, and I was not sure how people could end up in this situation, so could not have a better message. Can you point to a package and a reprex that produces this error?

3. We should also print out the name of the package, and the source file, if we can.

jameslamb:
> 1. I don't know how many people install packages from the command line. I would think most people use an IDE or `install.packages()`. On Windows, people probably don't use the shell much. So mentioning that option might be confusing.

Fair point. You can pass this to `INSTALL_OPTS` in an `install.packages()` call though and it  will get passed through to `R CMD INSTALL`.

> 2. roxygen does not use the installed package by default but it uses its own package code loading mechanism. Are you documenting the installed package? The reason for the current minimal package is that it is a safeguard, and I was not sure how people could end up in this situation, so could not have a better message. Can you point to a package and a reprex that produces this error?

I have been trying to create a reproducible example, let me see if I can come up with one. I hit this error  while working on [microsoft/LightGBM](https://github.com/microsoft/LightGBM/tree/master/R-package).  Because of our installation process there, I usually have to manually install the R package and then run  `roxygen2::roxygenize(load = 'installed')`. When I don't do that, it tries to recompile the C++ code in the package and I had a hard time control the values of e.g. `CXXFLAGS`.

Let me see if I can make a reproducible example.

> 3. We should also print out the name of the package, and the source file, if we can.

Could you advise me on how to do this? I had the same thought, but I don't actually understand what `x` passed in to `extract_r6_methods()` is. That would require a lot more knowledge of `roxygen2`'s internals.

gaborcsardi:
> Because of our installation process there, I usually have to manually install the R package and then run roxygen2::roxygenize(load = 'installed'). When I don't do that, it tries to recompile the C++ code in the package and I had a hard time control the values of e.g. CXXFLAGS.

> Let me see if I can make a reproducible example.

This should be enough, thanks. I suspected that it was with a different load method.

jameslamb:
> > Because of our installation process there, I usually have to manually install the R package and then run roxygen2::roxygenize(load = 'installed'). When I don't do that, it tries to recompile the C++ code in the package and I had a hard time control the values of e.g. CXXFLAGS.
> 
> > Let me see if I can make a reproducible example.
> 
> This should be enough, thanks. I suspected that it was with a different load method.

Alright sounds good. I'm really unsure how to make a reprex. In a super simple R package with  one un-exported R6 object, I tried this and couldn't get the error:

```shelll
R CMD INSTALL --without-keep.source .
Rscript -e "roxygen2::roxygenize(load = 'installed')"
```

gaborcsardi:
Thanks again! I improved the error message a bit, added the class name if available, and explained why this might happen. I'll merge after the CI runs are done.

gaborcsardi:
The CI failures are unrelated. @hadley Is this ok?

hadley:
@gaborcsardi yeah looks good to me.

gaborcsardi:
Thanks!

PR #1024, opened 2020-01-07
Support inline code

### LINKED ISSUE #1017: Implement inline R code
e.g. 

```R
#' `r title("my title")`
```

would be evaluated at document-time.

gaborcsardi:
It still needs docs and test.

Closes #1017.


gaborcsardi:
@hadley OK, it has tests, I think it is ready to review, and I can add some docs as well, in the formatting vignette. 

Some notes:

- The code loading now happens before the parsing. This can cause a change in behavior if one of them error. But this is probably fine.
- There is a new evaluation env for each tag. Within a tag, all code uses the same env. Is this right?
- If there is no package env, then a child of `baseenv()` is used, to facilitate testing. 
- Multi-line expressions currently error because cmark is buggy for these and messes up the locations in the text.
- We error on parse or eval errors (or other errors, e.g. multi-line messages), with a nice-ish message that is similar to the `roxy_warning()` format.
- We currently call `paste(, collapse = "\n")` on the value of the expression. So if you return `NULL`, you can run code, but no text is inserted.

gaborcsardi:
@hadley Added the docs as well.

tests/testthat/test-markdown-code.R
```diff
@@ -0,0 +1,145 @@
...
+  expect_equal(out1$get_value("field"), c(field = "Field 28"))
+})
+
+test_that("can see the package env", {
+  # This is better tested in the internal `markdown_pass` manual
+  # page, since there is a package env there, unlike here...
+  expect_true(TRUE)
+})
+
+test_that("uses the same env for a tag, but does not reuse envs", {
```
hadley: Should we share environments across all calls in one topic? (Or is that very hard? I think it is because this is all done at the tag level?)

tests/testthat/test-markdown-code.R
```diff
@@ -0,0 +1,145 @@
...
+    #'
+    #' Description `r exists('foobarxxx123', inherits = FALSE)`
+    #' @md
+    #' @name dummy
+    NULL")[[1]]
+  expect_equal(out1$get_value("title"), "Title 420 420")
+  expect_equal(out1$get_value("description"), "Description FALSE")
+})
+
+test_that("can create markdown markdup", {
```
hadley:
```suggestion
test_that("can create markdown markup", {
```

tests/testthat/test-markdown-code.R
```diff
@@ -0,0 +1,145 @@
...
+  expect_equal(out1$get_value("source"), "Source 22")
+  expect_equal(out1$get_value("param"), c(param = "Param 24"))
+  expect_equal(out1$get_value("slot"), c(slot = "Slot 26"))
+  expect_equal(out1$get_value("field"), c(field = "Field 28"))
+})
+
+test_that("can see the package env", {
+  # This is better tested in the internal `markdown_pass` manual
+  # page, since there is a package env there, unlike here...
+  expect_true(TRUE)
```
hadley: So delete this test?

tests/testthat/test-markdown-code.R
```diff
@@ -0,0 +1,145 @@
...
+    NULL")[[1]]
+  expect_equal(out1$get_value("title"), "Title 420 420")
+  expect_equal(out1$get_value("description"), "Description FALSE")
+})
+
+test_that("can create markdown markdup", {
+  out1 <- roc_proc_text(rd_roclet(), "
+    #' Title
+    #'
+    #' Description `r paste0('_', 'keyword', '_')`
```
hadley: I think the next few tests could just call `markdown()`?

tests/testthat/test-markdown-code.R
```diff
@@ -0,0 +1,145 @@
...
+      #'   1`--
+      #' @md
+      #' @name dummy
+      NULL")[[1]],
+    "in inline code: multi-line `r ` markup is not supported"
+  )
+})
+
+test_that("evaluation errors are reported nicely", {
+  expect_error(
```
hadley: Combine with the next test and use `verify_output()`?

vignettes/rd-formatting.Rmd
```diff
@@ -263,6 +265,73 @@ Markdown syntax for inline images works. The image files must be in the `man/fig
 #' ![](example-plot.jpg "Example Plot Title")
 ```
 
+# Dynamic R code
+
+Similarly to the knitr package, you can use the markdown inline code markup to evaluate R code and insert its output into the manual page.
+To do this, prefix the code with `r `, i.e. the lowercase letter 'r' and a space character.
+Roxygen will interpret the rest of the text within backticks as R code and evaluate it, and replace the backtick expression with its value.
```
hadley: , which will be interpreted as markdown.

R/markdown.R
```diff
@@ -1,18 +1,124 @@
...
+#' For example this becomes two: `r 1+1`.
+#' Variables can be set and then reused, within the same
+#' tag: `r x <- 100; NULL`
+#' The value of `x` is `r x`.
+#'
+#' We have access to the internal functions of the package, e.g.
+#' since this is _roxygen2_, we can refer to the internal `markdown`
+#' function, and this is `TRUE`: `r is.function(markdown)`.
+#'
+#' Roxygen created this manual page using R version `r getRversion()`.
```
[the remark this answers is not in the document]
gaborcsardi: Right. Removed.

tests/testthat/test-markdown-code.R
```diff
@@ -0,0 +1,145 @@
+
+test_that("can eval in various tags", {
+  out1 <- roc_proc_text(rd_roclet(), "
+    #' @title Title `r 1 + 1`
+    #'
+    #' @description Description `r 2 + 2`
```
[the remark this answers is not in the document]
gaborcsardi: Well, not all tags are parsed as markdown, so these are the tags that have it. But we already have such a list in the markdown tests, so yeah, I'll trim this down.

tests/testthat/test-markdown-code.R
gaborcsardi: It is definitely harder, and since tags can be inherited, etc. it makes everything complicated.

tests/testthat/test-markdown-code.R
gaborcsardi: yeah.

tests/testthat/test-markdown-code.R
gaborcsardi: I don't really want to rely on how the errors are printed, though.

vignettes/rd-formatting.Rmd
gaborcsardi: That would be a bit misleading, because the inserted value is not parsed as markdown, so I added this:
> After all such substitutaions, the text of the whole tag is interpreted as markdown, as usual.

OK?

hadley: :+1:

tests/testthat/test-markdown-code.R
hadley: But you're already depending on the exact text of the error?

gaborcsardi: Right, but if e.g. if we switch to rlang errors, it will be printed differently, and we might tweak the printout in different rlang versions, etc.

hadley: Right, so it's more robust to use `verify_output()` because that won't fail on CRAN?

gaborcsardi: Right, I did not realize that it does not fail on CRAN. Does it fail on the CI? IDK, I tend not to use `verify_output()` for non-output stuff. Also, I thought matching the error message should be fine, I assume that rlang would not mangle it. 

Anyway, I really don't mind very much, I can also use `verify_output()`.

tests/testthat/test-markdown-code.R
gaborcsardi: Right, done.

tests/testthat/test-markdown-code.R
gaborcsardi: I mean, not just the implementation would be more complicated, but I think it is harder to follow what exactly is evaluated.

tests/testthat/test-markdown-code.R
gaborcsardi: Done.

gaborcsardi:
@hadley All done I believe, except for deciding whether we do this at the tag level or elsewhere.

Doing it at the block level is easy as well, but doing it at the topic level is very hard imo. We would need to postpone the markdown parsing until the very end, which might not be a bad idea in itself, but it is not how roxygen works currently.

tests/testthat/test-markdown-code.R
gaborcsardi: done

hadley:
I think block level would perfect if it's not too hard.

gaborcsardi:
> I think block level would perfect if it's not too hard.

OK, it is block level now, I think it is pretty good, and intuitive.

I temporarily put the eval env into roxy_meta, which is kind of a hack, I can also just put it somewhere else within the package...

gaborcsardi:
@hadley Thanks much!

PR #1028, opened 2020-01-12
Support full ORCID URL in DESCRIPTION files

mikemahoney218:
As of a9c384e , ORCIDs have been formatted via the assumption that the DESCRIPTION field only contains the 16 digit code, rather than the full URL. Since the ORCID site seems to promote using the full URL, it seems like it could be useful to handle packages using the full URL as well.


hadley:
Fixes #1040

hadley:
Thanks @mikemahoney218!

PR #1044, opened 2020-02-10
Refactor writing Win/Unix line endings

### LINKED ISSUE #989: .Rd files are forcibly rewritten with LF ending on Windows
despite of what's announced in the NEWS and in #841.

To reproduce:

- Clone any package on Windows
- Run `unix2dos man/*.Rd`
- Run `devtools::document()`

I believe this is because `write_if_different()` already receives character strings with embedded newlines, and `write_lines()` doesn't convert them to the detected newlines.

gaborcsardi:
We need to move the detection of line endings
upstream a bit, to be able to tell if the
newly generated contents is different from the
current one, in `write_if_different()`.

This means that we need to `paste()` the contents
together in `write_if_different()`. Consequently,
we might as well remove the `paste()`-ing from
`same_contents()`, as `write_if_different()` always
calls it with the `paste()`-d contents.

Also, if we `paste()` already upstream, then
we might as well use `writeBin()` to write out
the contents, instead of `write_lines()`, especially
because `write_lines()` would add an extra empty line.
`write_lines()` is still kept, and used in some
code paths, and it can also do EOL-detection,
although this feature is not used currently.

Another benefit of this refactoring is that we
also move `enc2utf8()` upstream, this should clearly
happen before we check if the new contents is
different.

Added test cases as well.

Fixes #989, supersedes #1003.

### DIFF
--- a/R/utils-io.R
+++ b/R/utils-io.R
@@ -5,15 +5,7 @@ read_lines <- function(path, n = -1L) {
   base::readLines(path, n = n, encoding = "UTF-8", warn = FALSE)
 }
 
-write_lines <- function(text, path, line_ending = NULL) {
-  if (is.null(line_ending)) {
-    if (has_windows_le(path)) {
-      line_ending <- "\r\n"
-    } else {
-      line_ending <- "\n"
-    }
-  }
-
+write_lines <- function(text, path, line_ending = detect_line_ending(path)) {
   # we need to convert any embedded newlines as well
   text <- gsub("\r?\n", line_ending, text)
 
@@ -22,9 +14,9 @@ write_lines <- function(text, path, line_ending = NULL) {
   close(path)
 }
 
-has_windows_le <- function(path) {
-  tryCatch(
-    expr = isTRUE(grepl("\r\n", suppressWarnings(readChar(path, nchars = 500)))),
-    error = function(e) FALSE
-  )
+detect_line_ending <- function(path) {
+  tryCatch({
+    samp <- suppressWarnings(readChar(path, nchars = 500))
+    if (isTRUE(grepl("\r\n", samp))) "\r\n" else "\n"
+  }, error = function(e) "\n")
 }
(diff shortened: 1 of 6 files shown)


R/utils-io.R
```diff
@@ -5,13 +5,18 @@ read_lines <- function(path, n = -1L) {
...
 }
 
-write_lines <- function(text, path) {
-  if (has_windows_le(path)) {
-    line_ending <- "\r\n"
-  } else {
-    line_ending <- "\n"
+write_lines <- function(text, path, line_ending = NULL) {
+  if (is.null(line_ending)) {
+    if (has_windows_le(path)) {
```
krlmlr: Extract `get_le_from_path()` ?

R/utils.R
```diff
@@ -73,6 +73,14 @@ write_if_different <- function(path, contents, check = TRUE) {
...
   }
 
+  if (has_windows_le(path)) {
+    line_ending <- "\r\n"
+  } else {
+    line_ending <- "\n"
+  }
+
+  contents <- paste0(paste0(contents, collapse = line_ending), line_ending)
+  contents <- gsub("\r?\n", line_ending, contents)
```
krlmlr:
```suggestion
  contents <- enc2utf8(gsub("\r?\n", line_ending, contents))
```

and remove the conversion below?

gaborcsardi: Right, good catch, fixed now.

R/utils-io.R
gaborcsardi: That's an existing function, so I don't want to rename it.

gaborcsardi:
@krlmlr Thanks!

gaborcsardi:
> Maybe we should have a function that returns the line ending instead of a function that returns a logical if it’s a windows line ending? I think that would reduce a little duplication. And maybe make same_contents error of contents isn’t length 1?

I'll do these, I was considering them as well.

gaborcsardi:
@hadley OK, I did both, and added NEWS. I'll wait for the CI checks, and then merge.

PR #1045, opened 2020-02-13
Prefer markdown throughout the vignette

michaelquinn32:
This is more in line with the tidyverse style guide.


gaborcsardi:
Good idea! I am not sure about the `\describe{},` though, because the markdown is not really equivalent.

michaelquinn32:
Thanks for the feedback!

Maybe others might be interested in chiming in, but I believe:

* Having one way of doing things, especially if it follows the style guide, is always better
* It doesn't really matter that the markdown to Rd conversion matches describe; it's close enough
* The markdown version is more readable anyway, diminishing the importance of the output further

I would imagine that the last part is a bit contentious. Is the true target for the documentation the code itself or another format (like what we call from interactive R)? At Google, we mostly care about the inline documentation, and the generated Rd files are of secondary importance. I would be happy to sacrifice `\describe{}` if I can get a close-enough markdown equivalent.

gaborcsardi:
Yeah, I don't have any hard data, but strongly suspect that most R users do not read the docs from the code of the package, but either via the (HTML) help browser, or the (different HTML) pkgdown web site.

michaelquinn32:
Here's a use of `describe`:
https://rdrr.io/cran/psych/man/Gleser.html

This is pretty close to a markdown list without bullets. If I add formatting to those items, would that be close enough?

gaborcsardi:
This is in the terminal, first the markdown list, then the `\describe{}`:

```
        • ‘foo’: Often status messages are associated with processes.
          E.g. the app starts downloading a large file, so it sets the
          status bar accordingly. Once the

        • ‘bar’: short one

        • another one: Often status messages are associated with
          processes. E.g. the app starts downloading a large file, so
          it sets the status bar accordingly. Once the

     ‘foo’: Often status messages are associated with processes. E.g.
          the app starts downloading a large file, so it sets the
          status bar accordingly. Once the

     ‘bar’: short one

     ‘another one’: Often status messages are associated with
          processes. E.g. the app starts downloading a large file, so
          it sets the status bar accordingly. Once the
```

And this is in the HTML help, in RStudio:

Screen Shot 2020-02-26 at 00 35 25

The `\describe{}` output is better, I think. So I am not sure if we should change this in the manual just yet. Ideally, we would have a way to write a description list in commonmark markdown.

michaelquinn32:
Fair enough. I can revert that change so the section stuff can be checked in.

hadley:
Thanks @michaelquinn32 !

PR #1053, opened 2020-03-06
Fix object format test case for R 4.x

### LINKED ISSUE #1052: Fix test case for object format, "matrix" inherits from "array", starting from R 4.0.0 (current R-devel)
```
-- 1. Failure: format has nice defaults for bare vectors (@test-object-format.R#
Results have changed from known value recorded in 'test-object-format.txt'.
1/15 mismatches
x[8]: "[1] \"An object of class \\\\code{matrix} (inherits from \\\\code{array})
x[8]:  with 10 rows and 10 columns.\""
y[8]: "[1] \"An object of class \\\\code{matrix} with 10 rows and 10 columns.\""
```

https://github.com/r-lib/roxygen2/runs/489923441?check_suite_focus=true

gaborcsardi:
Matrices now inherit from "array", so that
is included in the output.

Closes #1052.

### DIFF
--- a/tests/testthat/test-object-format-r4.txt
+++ b/tests/testthat/test-object-format-r4.txt
@@ -1,0 +1,15 @@
+> call_to_format(x <- list(a = 1, b = 2))
+[1] "An object of class \\code{list} of length 2."
+
+> call_to_format(x <- ordered(letters[1:5]))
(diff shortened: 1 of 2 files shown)


gaborcsardi:
@hadley There are other ways to fix this, we could also just omit the test for matrix.

gaborcsardi:
Thanks!

Issue #988, opened 2019-11-25
Include Rmd in an user defined section


maelle:
The new includeRmd feature is awesome, thanks! :star2: 

I read "Example: @includeRmd path description. This currently does not work with user defined sections (created with @section)." in [the vignette](https://roxygen2.r-lib.org/articles/rd.html#including-external--rmd-md-files) and am therefore opening an issue to request the feature. :wink:

I would like to use it in particular to be able to include content _after_ the "Value" and "Examples" section without needing to write a level 1 header in the Rmd. The Details section comes before "Value" and "Examples" which is not optimal for the user reading the docs.

Not sure if my use case is helpful to know about: I have an Rmd with all the details the user needs to know about authorization in a package, whose content is under a level 3 header in the README and vignette (via [the approach described in this blog post](https://www.garrickadenbuie.com/blog/dry-vignette-and-readme/)), but should be an user-defined section in the man page. At the moment I can't do that without having a level 1 header in the Rmd, that I'd somehow strip in a previous chunk in the README and vignette.

hadley:
I don't think you can insert content after the examples section.

gaborcsardi:
I'll take a look at this before the release. To include `.Rmd`s into user defined `@section`s I mean. 

The order of the sections in the formatted manual is beyond what we can do.

gaborcsardi:
Yeah, so this is not so easy to fix, because roxygen tags cannot be embedded. The `@includeRmd` tag would have to know which `@section` tag came before it, which is probably possible to implement, but it seems messy.

I think with the new markdown code support it will be much easier to solve these issues. E.g.

```
#' @section Title:
#' Some text.
#'
#' ```{r child = "man/roxygen/child.Rmd"}
#' ```
#'
#' More text.
```

A child document like this will be always embedded into the current tag.

PR #1055, opened 2020-03-06
Keep order of @includeRmd sections

### LINKED ISSUE #1032: Order of @includeRmd custom sections and pkgdown pages
When building a reference topic page, pkgdown lays out sections according to the order found in the original rd file. Currently `@includeRmd` places custom sections at the bottom of an rd file if the section is not a part of an existing section. In my example, the custom section included appears after the "see also" section (alternatively the "see also" section comes before the custom section).

image

I think a solution for #988 would nullify this problem. Custom sections could be manually placed and the contents of include rmd would be added to the existing sections.

gaborcsardi:
Use 'section' sections for level 1 headings, instead of a 'rawRd' sections.

The return values of the internal `markdown()` and `mdxml_children_to_rd_top()` can now be named vectors, where the names are the section titles. Not super great, but it is was already not very clean to start with.

I'll take another look at this on Monday, and will try to clean up this part a bit. Will also try to address #988 then.

Closes #1032.

### DIFF
--- a/tests/testthat/test-markdown.R
+++ b/tests/testthat/test-markdown.R
@@ -570,19 +570,12 @@ test_that("level 1 heading in @details", {
     #' Description.
     #' @details
     #' Leading text goes into details.
-    #' @rawRd
-    #' \\section{This is its own section}{
-    #' \\subsection{Can have a subsection}{
+    #' @section This is its own section:\\subsection{Can have a subsection}{
     #'
     #' Yes.
     #' }
     #'
-    #' }
-    #' @rawRd
-    #' \\section{Another section}{
-    #'
-    #' With text.
-    #' }
+    #' @section Another section:With text.
     #' @name x
     NULL
   "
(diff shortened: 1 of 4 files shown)


tests/testthat/test-markdown.R
```diff
@@ -570,19 +570,12 @@ test_that("level 1 heading in @details", {
     #' Description.
     #' @details
     #' Leading text goes into details.
-    #' @rawRd
-    #' \\section{This is its own section}{
-    #' \\subsection{Can have a subsection}{
+    #' @section This is its own section:\\subsection{Can have a subsection}{
```
hadley: Can you put a newline after the `:`?

gaborcsardi: Well, if I want to match the output generated by the markdown, then I can't.  Or we can trim / simplify the output before comparison.

hadley: Ah ok

gaborcsardi: OK, I'll merge this now then, and will see if #988 is possible without too much refactoring.

PR #1059, opened 2020-03-10
Add new packgaes meta field

### LINKED ISSUE #1013: Allow user-defined tags from external packages, have a way to load an external package in `roxygenize()`
Before:

```yaml
Roxygen: { library(roxygenlabs); list(markdown = TRUE,
    roclets = c("collate", "namespace", "roxygenlabs_rd")) }
```

After:

```yaml
Roxygen: list(
      markdown = TRUE,
      roclets = c("collate", "namespace", "roxygenlabs::roxygenlabs_rd")
    )
```

hadley:
Not sure how to test this, but hopefully it's sufficiently simple I didn't mess it up.

Fixes #1013

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # roxygen2 (development version)
 
+* Roxygen metadata can now have a `packages` element, giving a character vector 
+  of package names to load. This makes it easier to use extension package that 
+  provide new tags for existing roclets (#1013).
+  
+    ```yaml
+    Roxygen: list(markdown = TRUE, packages = "roxygenlabs")
+    ```
+  
 * The warning for section titles spanning multiple lines now includes the hint 
   "Did you forget a colon (:) at the end of the title?" (@maelle, #994).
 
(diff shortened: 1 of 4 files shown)


hadley:
@gaborcsardi `requireNamespace()` already errors if the package isn't found?

gaborcsardi:
> @gaborcsardi requireNamespace() already errors if the package isn't found?

@hadley But you call it with `quietly = TRUE`, in which case it does not error:
https://github.com/r-lib/roxygen2/pull/1059/files#diff-5536902b32f437aa37fb6527dece2f7cR45

```r
❯ requireNamespace("foobarsdfsdf", quietly = TRUE)
❯
```

Actually, even with `quietly = FALSE` it does not error, just prints an error message:

```r
❯ { requireNamespace("foobarsdfsdf", quietly = FALSE); print("still here") }
Loading required namespace: foobarsdfsdf
Failed with error:  ‘there is no package called ‘foobarsdfsdf’’
[1] "still here"
```

gaborcsardi:
On the bright side, I tried it on a package, and it does work nicely.

hadley:
Oh, maybe I should by using `loadNamespace()`?

gaborcsardi:
Yeah, that's good as well, that will error for sure.

PR #1073, opened 2020-04-08
Missing a line from `str_length()`

rorynolan:


hadley:
The point here is to show the documentation, not the implementation of `str_length()`, so I'd prefer to leave it as is.

PR #1078, opened 2020-04-22
Fix `rd_family_title` option

salim-b:
Support for custom `@family` titles was added in #883 and [vignettes/rd.Rmd](https://roxygen2.r-lib.org/articles/rd.html) explains how to use them (explanation was added in 0224086bf189094e87fec44aafe694a79f23ffe7).

But

1. The `man/roxygen/meta.R` example in the vignette is wrong (missing list level).

2. roxygen2 currently doesn't recognize the `rd_family_title` option (building the doc throws the warning `Unknown Roxygen options rd_family_title.`).

There's already [an unresolved question on Stackoverflow](https://stackoverflow.com/questions/60493979/r-documentation-the-family-tag-is-not-working) regarding this.

This PR adds the missing pieces, so `rd_family_title` works as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 *.so
-.RData
 .Rproj.user
 /.Rhistory
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.so
+.RData
 .Rproj.user
 /.Rhistory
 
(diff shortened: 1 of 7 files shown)


salim-b:
Any update on this?

gaborcsardi:
Thanks! We'll look at this soon, when we'll work on roxygen2 again.

hadley:
Thanks!

PR #1109, opened 2020-06-11
Link to files, not topics

### LINKED ISSUE #1108: Markdown links need to link to the file instead of the topic now
Following the change in https://github.com/wch/r-source/commit/f27cbf1a52a31cd9b9676340394946a22041a4ae

This was discussed in #707, but the situation is different now, with the check added.

gaborcsardi:
External packages are simple, we just look up the file name and link to that. 

~~Linking to the package being documented is harder, because at the time of the linking we don't know the file names yet. So we'll need to mark these parts in the docs, and post-process all files, ones the file names are known.~~ These links do work with the topic, so we'll leave them alone.

- [x] Links to external packages.
- [x] ~~Internal links, to the package being documented.~~ Not needed.
- [x] Fix re-exported functions, if they need anything special
- [x] Improve warnings to contain file name and line number.
- [x] Throw warnings for non-existing package or topic, for re-exports. 
- [x] Simplify links where the file name is the same as the topic name.
- [x] Refactor into its own file.
- [x] ~~Can we link to another package w/o specifying its name? Yes, probably shouldn't, though.~~ Not relevant.
- [x] ~~Should we look up unqualified unknown link targets in dependent packages? Not just yet.~~ Not releavant.
- [x] ~~One warning only for non-local unqualified links.~~ No warning at all, `R CMD check` will warn if it needs to.
- [x] ~~Resolve links to the base package, automatically?~~ No need, we'll just leave unqualified links alone.
- [x] Tests.
- [x] ~~Docs.~~ Not really needed.
- [x] News.
- [x] Fix links in inherited text.
- [x] Fix links in the text of the `@inheritedDotParams` intro.
- [x] Automatically remove package names from in-package references. This is e.g. useful for compat files that are copied to a number of packages. It does not need to be in this PR, just making a note here.

Closes #1108.


gaborcsardi:
OK, this PR has an implementation now. The biggest glitch is that it always includes the file name for internal links (i.e. links within the documented package), even if it is the same as the topic name.

gaborcsardi:
> - [x] Can we link to another package w/o specifying its name?

Yes. The link will be resolved dynamically, using the packages that are loaded at the time. This is problematic:

* If the package is not loaded, then R does not find the page, or finds a different page with the same name or alias.
* If multiple packages define the same topic, then the user will have to choose between them, and this is not always an informed choice.

> - [ ] Should we look up unqualified unknown link targets in dependent packages?

If we don't do that, then we are prone to the problems just above. Ideally we should also consider the dependency tree.

OTOH, the current version of roxygen2 does not do anything special about this, so we can do this later.

For now, maybe we can have a warning that lists all references that were not qualified?

gaborcsardi:
OK, complete change of plans now. The only problematic links style that roxygen2 creates is `\link[pkg:topic]{text}`. This is created from the following expressions:
* `[pkg::func()]`
* `[pkg::thing]`
* ``[`pkg::thing`]``
* `[text][pkg::func()]`
* `[text][pkg::thing]`

We can change the Rd output for this from 
```
\link[pkg:topic]{text}
```
to
```
\ifelse{html}{\link[pkg:topic]{text}}{\link[=topic]{text}}
```
This satisfies the check and still links to the topic instead of the file name. If the link should be typeset as code, then we need to write
```
\ifelse{html}{\code{\link[pkg:topic]{text}}}{\code{\link[=topic]{text}}}
```

yutannihilation:
Thanks for implementing this so quickly! I have two questions after trying this on ggplot2.

1. I see a lot of warnings on unqualified function names. For now, do I need to specify all of them even if it's `base::`?
2. It seems the escape for `%` is not considered when roxygen2 resolves the reference. I see the warning below. Do I need to fix some notation?

```
Warning: [theme.r:168] Link to unknown topic '\%+replace\%'
```

original doc:

``` r
#'   [+.gg()] and [%+replace%],
```

gaborcsardi:
> I see a lot of warnings on unqualified function names. For now, do I need to specify all of them even if it's base::?

No, you can ignore them for now. The links themselves work fine.

> It seems the escape for % is not considered when roxygen2 resolves the reference. I see the warning below. Do I need to fix some notation?

If the link itself works, then just ignore this.

You might want to wait a day, because I'll solve this issue completely differently and that solution will generate different Rd files: https://github.com/r-lib/roxygen2/pull/1109#issuecomment-643607435

yutannihilation:
Ah, thanks, I'll wait. Sorry, I took your comment differently and thought the last commit implemented it...

gaborcsardi:
Actually, the approach in https://github.com/r-lib/roxygen2/pull/1109#issuecomment-643607435 does not work, so we'll have to go with the more complicated solution.

gaborcsardi:
@yutannihilation If I run the current version on gghighlight, I only get a single change in a link to ggrepel, an no warnings at all.

gaborcsardi:
@hadley This is ready, fyi.

yutannihilation:
Thanks, I confirmed I don't see warnings on `devtools::document()`! However, I see these warnings on R CMD check with R-devel:

```
* checking Rd cross-references ... WARNING
Non-file package-anchored link(s) in documentation object 'as.list.ggproto.Rd':
  ‘[base:as.list.environment]{base::as.list.environment}’

Non-file package-anchored link(s) in documentation object 'cut_interval.Rd':
  ‘[base:cut.default]{base::cut.default}’

Non-file package-anchored link(s) in documentation object 'scale_brewer.Rd':
  ‘[RColorBrewer:brewer.pal]{RColorBrewer::brewer.pal()}’

Non-file package-anchored link(s) in documentation object 'tidyeval.Rd':
  ‘[rlang]{syms}’ ‘[rlang]{expr}’ ‘[rlang]{quo}’ ‘[rlang]{exprs}’
  ‘[rlang]{quos}’ ‘[rlang]{enexpr}’ ‘[rlang]{enquo}’ ‘[rlang]{enexprs}’
  ‘[rlang]{enquos}’
```
(https://travis-ci.org/github/tidyverse/ggplot2/jobs/698111831#L1540-L1553)

First and second ones are something related to `@inheritDotParams`, and third one related to `@inheritParams`. I have no idea about the last one.

gaborcsardi:
Yes, if you are inheriting docs from another package, that needs to be fixed in that package. One solution is to download the CRAN version of that package, run `document()` on it with this patch, and then install it. Then re-run `document()` on the inheriting package.

This is certainly not a joy if you are inheriting a lot of arguments or code, so it is possible that we need to do something about it.

yutannihilation:
Wow, cool! (For ggplot2, it's only scales package that I needed to tweak, so it wouldn't be a heavy burden even if I need to do it manually).

For `@inheritDotParams`, however, it doesn't seem the case where inherited docs are the problem. Can this PR also fix the link in the phrase below?

```
\item{...}{
  Arguments passed on to \code{\link[base:as.list.environment]{base::as.list.environment}}
```

gaborcsardi:
Yes, it is not in the inherited text, but in the text we add as explanation. Good point. Sure, I should be able to fix that as well.

gaborcsardi:
@yutannihilation OK, all fixed now, I believe. The rlang related warnings need to be fixed in ggplot itself, as you mentioned above.

yutannihilation:
Confirmed this fixes all the warnings (except for rlang ones) now! Thanks so much!

krlmlr:
Does this mean that the *file names* in the `man/` directory are now considered part of a package's API? :zany_face:

R/object-import.R
```diff
@@ -34,3 +39,19 @@ format.rd_section_reexport <- function(x, ...) {
     "\n}}\n"
   )
 }
+
+find_topic_in_package_reexp <- function(pkg, topic) {
+  path <- tryCatch(
```
hadley, line 44: It feels like this pattern is repeated in a couple of places, so might be worth extracting out.

R/markdown-link.R
```diff
@@ -111,6 +114,8 @@ parse_link <- function(destination, contents, state) {
   ## `obj` is fun or obj (fun is without parens)
   ## `s4` is TRUE if we link to an S4 class (i.e. have -class suffix)
   ## `noclass` is fun with -class removed
+  ## `file` is the file name of the linked topic, if known. Otherwise random
```
hadley, line 117: This comment is out of date now, right?

R/object-import.R
```diff
@@ -19,7 +19,12 @@ merge.rd_section_reexport <- function(x, y, ...) {
 format.rd_section_reexport <- function(x, ...) {
   pkgs <- split(x$value$fun, x$value$pkg)
   pkg_links <- map2(names(pkgs), pkgs, function(pkg, funs) {
-    links <- paste0("\\code{\\link[", pkg, "]{", escape(sort(funs)), "}}",
+    funs <- sort(funs)
+    files <- vapply(funs, find_topic_in_package_reexp, character(1), pkg = pkg)
+    links <- paste0(
+      "\\code{\\link[", pkg,
+      ifelse(files == funs, "", paste0(":", files)),
```
hadley, line 26: I _think_ it be simpler for `find_topic_in_package_reexp()` to do the pasting together of package and file?

R/rd-find-link-files.R
```diff
@@ -0,0 +1,63 @@
...
+#' topic alone.
+#'
+#' If `pkg` is not `NA` and not the package being documented (the _dev_
+#' package), then we need to be able to find the Rd file. If we can't, that's
+#' a warning and the link is left untouched. This typically happens when the
+#' linked package is not installed or cannot be loaded.
+#'
+#' @noRd
+
+find_topic_filename <- function(pkg, topic, tag) {
```
hadley, line 23:
```suggestion
find_topic_filename <- function(pkg, topic, tag = NULL) {
```

R/object-import.R
```diff
@@ -34,3 +39,19 @@ format.rd_section_reexport <- function(x, ...) {
...
+    error = function(err) {
+      roxy_warning("Unavailable package in re-export: ", pkg, "::", topic)
+      topic
+    }
+  )
+  if (is.na(path)) {
+    roxy_warning("Unavailable topic in re-export: ", pkg, "::", topic)
+    topic
+  } else {
+    path
```
hadley, line 55: i.e. should this be `paste(pkg, ":", topic)`?

tests/testthat/test-markdown-link.R
```diff
@@ -94,16 +94,19 @@ test_that("short and sweet links work", {
     foo <- function() {}")[[1]]
   expect_equivalent_rd(out1, out2)
 
-  out1 <- roc_proc_text(rd_roclet(), "
+  expect_warning(
+    out1 <- roc_proc_text(rd_roclet(), "
     #' Title
     #'
-    #' See [pkg::function()], [pkg::object].
+    #' See [11pkg::function()], [11pkg::object].
```
hadley, line 101: Would it be better to replace these with actual functions/objects?

cboettig:
Not entirely sure if this is related, but in R-devel I now see my roxygen code links to S4 class definition pages breaking where it was fine in the current release?  e.g. I use ` \code{\link[XML]{XMLInternalNode-class}}`, which has been fine in all previous versions.  Apologies if this is not a roxygen issue, but I'm unclear in how to link to the documentation you get by 

```
 ?`XML::XMLInternalNode-class` 
```

in either the standard or the markdown syntax in a way that makes R-devel checks happy...

gaborcsardi:
@cboettig If you use `\link[]{}`, roxygen2 does not touch that (except in another topic that inherits this). So it is not a roxygen2 issue. You can look up the name of the file like this:

```r
> help("XMLInternalNode-class", "XML")[[1]]
[1] "C:/Users/csard/R/win-library/4.0/XML/help/XMLNode-class"
```

So you need something like `\code{\link[XML:XMLNode-class]{XMLInternalNode-class}}` or `\code{\link[XML:XMLNode-class]{XMLInternalNode}}`.

In roxygen2 you would only need `[XML::XMLInternalNode-class]` and roxygen2 does the rest for you.

R/markdown-link.R
gaborcsardi: Right, fixed.

R/object-import.R
gaborcsardi: Just tried, one part, yes. It makes the other part more cumbersome, as we cannot to `ifelse(files == funs, ...)` any more, would need to `ifelse(files == paste0(pkg, ":", funs), ...)` which is not terrible, but the `files == funs` part is/was quite expressive, i.e. we only do this if the file name is different than the topic name.

Also, all the other `find_topic_in_package*()` functions just return the topic, so I think we can just stick with that?

R/object-import.R
gaborcsardi: Yeah.

R/object-import.R
gaborcsardi: Maybe not, because then some callers need to take it apart again?

tests/testthat/test-markdown-link.R
gaborcsardi: Hmmm. This is the test for the warnings we get for non-existing packages and topic. So I wanted to make sure that they indeed don't exist.

gaborcsardi:
@hadley Can you pls take another look? Thanks!

.github/workflows/R-CMD-check.yaml
```diff
@@ -67,6 +67,7 @@ jobs:
         run: |
           remotes::install_deps(dependencies = TRUE)
           remotes::install_cran("rcmdcheck")
+          remotes::install_github("rstudio/rmarkdown")
```
hadley, line 70: I think you can remove this now 😄

R/object-import.R
hadley: Ok, sounds good.

R/rd-find-link-files.R
```diff
@@ -0,0 +1,90 @@
...
+#' @return File name if the topic was found, `NA` if the package could be
+#'   searched, but the topic was not found. Errors if the package cannot be
+#'   searched. (Because it is not installed or cannot be loaded, etc.)
+#'
+#' @noRd
+
+find_topic_in_package <- function(pkg, topic) {
+  # This is needed because we have the escaped text here, and parse_Rd will
+  # un-escape it properly.
+  raw_topic <- str_trim(tools::parse_Rd(textConnection(topic))[[1]][1])
```
hadley, line 48: Does this leave a dangling connection?

R/rd-find-link-files.R
```diff
@@ -0,0 +1,90 @@
...
+  basename(utils::help((raw_topic), (pkg))[1])
+}
+
+try_find_topic_in_package <- function(pkg, topic,
```
hadley, line 52: I think you could simplify still further so you could write:

```R
try_find_topic_in_package(pkg, topic, " in re-export")
```

And then you wouldn't need the two helper functions.

tests/testthat/test-markdown-link.R
hadley: Oooh, makes sense.

R/rd-find-link-files.R
gaborcsardi: I think read only text connections are gracefully closed, without a warning:
```r
> readLines(textConnection("foo\n"))
[1] "foo" ""   
> gc()
          used (Mb) gc trigger (Mb) max used (Mb)
Ncells  665032 35.6    1496105   80  1340143 71.6
Vcells 1122545  8.6    8388608   64  2094520 16.0
> gc()
          used (Mb) gc trigger (Mb) max used (Mb)
Ncells  665014 35.6    1496105   80  1340143 71.6
Vcells 1122532  8.6    8388608   64  2094520 16.0
> gc()
          used (Mb) gc trigger (Mb) max used (Mb)
Ncells  665014 35.6    1496105   80  1340143 71.6
Vcells 1122532  8.6    8388608   64  2094520 16.0
```

But I'll clean it up, just in case. It does not hurt.

gaborcsardi:
@hadley: Thanks, this is ready to go.

hadley:
Thanks!

PR #1116, opened 2020-06-15
Fix interactions of dynamic code and fragile tags

### LINKED ISSUE #1115: Use of inline HTML is not currently supported
The lifecycle badges produce correct output but with warnings:

```r
roxygen2::roc_proc_text(roxygen2::rd_roclet(), "
  #' Foo
  #' @name foo
  #' @md
  #' @description
  #' `r lifecycle::badge('stable')`
  NULL
")
#> Warning: [<text>:5] @description Use of inline HTML is not currently supported
#> Warning: [<text>:5] @description Use of inline HTML is not currently supported
#> Warning: [<text>:5] @description Use of inline HTML is not currently supported
#> $foo.Rd
#> \name{foo}
#> \alias{foo}
#> \title{Foo}
#> \description{
#> \ifelse{html}{\out{<a href='https://www.tidyverse.org/lifecycle/#stable'><img src='figures/lifecycle-stable.svg' alt='Stable lifecycle'></a>}}{\strong{Stable}}
#> }
```

Needed for https://github.com/r-lib/lifecycle/issues/17.

gaborcsardi:
Before this, we evaluated dynamic after the fragile tags
were removed (set aside temporarily). This is problematic,
because these tags were also removed from within inline
code and code blocks, and more importantly, you could
not reliably return fragile tags from the dynamic code.

After this patch the order is this:
1. Parse the tag as markdown, find the inline code
   and the code blocks, execute them, substitute
   back the result.
2. Set aside the fragile tags, temporarily.
3. Parse as markdown again, generate Rd.
4. Put back the fragile tags.

This is a breaking change in theory, however dynamic
code is relatively recent, and the affected documentation
is somewhat obscure, so we can probably get away with it.
The potentially affected cases are:
* People having dynamic code in fragile tags. E.g.

    #' blah \out{`r expression`} blah

  This will now evaluate the expression.
* People generating fragile tags dynamically. E.g. as in
  the original post of https://github.com/r-lib/roxygen2/issues/1115
  We actually improve this use case with this patch.

Closes #1115.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # roxygen2 (development version)
 
+* roxygen2 now supports inline markdown code and code chunks inside
+  Rd tags. In particular in `\out{}` (#1115).
+
 # roxygen2 7.1.0
 
 ## New features
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -1,5 +1,6 @@
 
 markdown <- function(text, tag = NULL, sections = FALSE) {
+  tag <- tag %||% list(file = NA, line = NA)
   tryCatch(
     expanded_text <- markdown_pass1(text),
     error = function(e) {
@@ -10,7 +11,8 @@ markdown <- function(text, tag = NULL, sections = FALSE) {
       stop(message, call. = FALSE)
     }
   )
-  markdown_pass2(expanded_text, tag = tag, sections = sections)
+  escaped_text <- escape_rd_for_md(expanded_text)
+  markdown_pass2(escaped_text, tag = tag, sections = sections)
 }
 
 #' Expand the embedded inline code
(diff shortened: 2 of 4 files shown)


gaborcsardi:
@hadley I am not sure if this needs a NEWS entry. Maybe a comment in the docs would be good.

gaborcsardi:
OK, added NEWS. 

I just realized, that with this change we could easily allow dynamic markdown code in roxygen tags that don't use markdown currently. In particular, it would be very useful in `@examples` I think. For these tags, we would only run the first pass of the current markdown parser. I.e. we would evaluate the dynamic code, but then this would have to produce proper Rd markup. I'll add an issue for this.

PR #1125, opened 2020-06-26
Use cpp11

jimhester:
This conversion was pretty straightforward. The main thing 'missing' was cpp11 does not support default arguments, so I had to explicitly add them to the various function calls. One could argue this is actually of benefit I guess.

I am still undecided if cpp11 should [handle default arguments](https://github.com/r-lib/cpp11/issues/4).


jimhester:
This seems fine apart from failing due to codecov patch, which seems mostly spurious to me, maybe we should just run `usethis::use_coverage()` here to remove that check.

hadley:
Thanks @jimhester!

Issue #1077, opened 2020-04-17
Items should be sorted with C locale


wch:
Sorting can depend on locale.

In the master branch of https://github.com/rstudio/shiny (currently at 12401b6), the sorting of items in man/reexports.Rd depends on value of `LC_COLLATE`.

The two sets of command below produce different results.

```R
Sys.setlocale("LC_COLLATE", "en_US.UTF-8")
devtools::document()
 
Sys.setlocale("LC_COLLATE", "C")
devtools::document()
```

Here's the diff. With en_US.UTF-8, `HTML` is between `hr` and `htmlTemplate`. With C, `HTML` comes before `a`.

```diff
$ git diff
diff --git a/man/reexports.Rd b/man/reexports.Rd
index 41386ac1..4aa35fe5 100644
--- a/man/reexports.Rd
+++ b/man/reexports.Rd
@@ -52,6 +52,6 @@ below to see their documentation.
 \describe{
   \item{fastmap}{\code{\link[fastmap]{is.key_missing}}, \code{\link[fastmap]{key_missing}}}
 
[...]

PR #1132, opened 2020-07-03
Use sort_c after ls, so RC methods have consistent order

d-sci:
Completes #1077, also for Reference Classes. There was the exact same issue, but it was less obvious because it was due to `ls` rather than `sort`. 

Concrete minimal example (when I originally prepared this I thought it was an RStudio build pane quirk, until I saw #1077):
```
#' @title Case Sensitivity Test
#' @description
#' When you run [devtools::document()] in the R console, the ordering of RC
#' methods is case insensitive. When you run documentation in the build pane
#' (e.g. via `Ctrl+Shift+D`), the ordering is case sensitive.
#' @export CaseSensitivity
#' @export
CaseSensitivity <- setRefClass("CaseSensitivity",
  methods = list(
    aa = function(){
      'If case insensitive, aa will come before AZ'
      return("insensitive")
    },
    AZ = function(){
      'If case sensitive, AZ will come before aa'
      return("sensitive")
    }
  ))
```

would lead to:
image


R/object-rc.R
```diff
@@ -24,7 +24,7 @@ rc_methods <- function(obj) {
   parent_methods <- unlist(lapply(parents, function(x) {
     methods::getRefClass(x)$methods()
   }))
-  method_names <- setdiff(ls(envir = obj@refMethods), parent_methods)
```
hadley, line 27: How about just using `sort_c()` here? That makes it very clear that the results are sorted.

d-sci: Makes sense! Updated

hadley:
Thanks!

PR #1146, opened 2020-08-08
This commit adds the following tags:

mamueller:
@s4generics, @s4methods, @s4superclasses, @s4subclasses can be used to
extend the documentation of generic fucntions, methods and classes by
information that can be inferred by inspection of the underlying objects
via the methods package.
I have included tests that serve also to demonstrate how the
documentation is enhanced if the new tags are present.
The 'units' are tiny Rpackages which is not unnatural, given that the
scope of the search for sub and super classes as well as the inherited
methods is the package.
The output is checked for cran compatibility but real (small scope) unit
tests are not provided yet.


hadley:
This a lot of code, and I don't think there's been any discussion about the motivation or goals. If you want to do a big PR, it's best to start with an issue where we can discuss whether or not the code is a good fit for the package.

Issue #1096, opened 2020-05-12
Update table summarizing when examples are run?


maelle:
Should [this handy table](https://roxygen2.r-lib.org/articles/rd.html#functions) be updated?

Cf https://github.com/r-lib/devtools/issues/2216#issuecomment-619707307 (h/t https://community.rstudio.com/t/r-cmd-check-r-4-0-0-now-runs-donttest-how-to-proceed-with-long-running-examples/64493)

gaborcsardi:
Maybe that table should have a column for CRAN incoming tests?

PR #1167, opened 2020-11-13
Update table summarizing when examples are run?

maelle:
#1096


maelle:
this comment would be useful to have in the docs too: https://github.com/r-lib/devtools/issues/2216#issuecomment-666577533

PR #1174, opened 2020-11-25
close connections in tests

MichaelChirico:


maxheld83:
I unfortunately created what I now see is a duplicate PR addressing the same problem #1189 in  #1191.
My bad, really should've checked here first 😞.

Can we combine these into one, to save the maintainers some hassle @MichaelChirico?

"Mine" in #1191 is a bit shorter, and it just leaves the whole handling of connections to base R. 

With your permission, can I add you as a co-committer to #1191 (since you fixed this first) and we close this?

MichaelChirico:
sure please do!

maxheld83:
great, thanks, done!

This can be closed then in favor of #1191

PR #1188, opened 2021-01-19
eval=F for one chunk in vignettes/rd.Rmd

mpadge:
Vignette [currently renders](https://roxygen2.r-lib.org/articles/rd.html#packages-1) like this:
``` r
#' @details
#' The only function you're likely to need from roxygen2 is [roxygenize()]. 
#' Otherwise refer to the vignettes to see how to format the documentation.
#' @keywords internal
"_PACKAGE"
#> [1] "_PACKAGE"
```
where the terminal `#> [1] "_PACKAGE"` comes from evaluating the chunk. This PR switches off evaluation to suppress that last line.


hadley:
Thanks!

Issue #1189, opened 2021-01-19
repeated test runs in same session may raise warnings


maxheld83:
As of https://github.com/r-lib/roxygen2/commit/b3053b694399416543af877837efb29e6ec34d1f, repeated tests with some failing tests in the same session may raise these warnings:

```
1: In .Internal(gc(verbose, reset, full)) :
  closing unused connection 7 (/var/folders/1f/9tm8hrm179b6qsv8mw9lbv580000gn/T//RtmpWHa5Ey/file92741d50bb5e)
2: In .Internal(gc(verbose, reset, full)) :
  closing unused connection 6 (/var/folders/1f/9tm8hrm179b6qsv8mw9lbv580000gn/T//RtmpWHa5Ey/file9274512de9cd)
3: In .Internal(gc(verbose, reset, full)) :
  closing unused connection 5 (/var/folders/1f/9tm8hrm179b6qsv8mw9lbv580000gn/T//RtmpWHa5Ey/file92746b389a99)
4: In .Internal(gc(verbose, reset, full)) :
  closing unused connection 4 (/var/folders/1f/9tm8hrm179b6qsv8mw9lbv580000gn/T//RtmpWHa5Ey/file92743e1b3129)
```

The exact number of closed unused connections varies somewhat.

To reproduce:

1. Introduce a failing test somewhere:
    ```r
     test_that("trigger unused connections warning",  testthat::fail())
    ```
2. Run the test at least twice:
    ```r
    devtools::test()
    devtools::test()
    ```

This isn't really a big problem, though it can get contributors off on the wrong path if their tests break during development.
(I thought I might must have somehow mishandled connections in https://github.com/r-lib/roxygen2/pull/1182).

Based on the paths, my guess would be that these warnings are caused by connections to `tempfile()`  in tests, which somehow may not be properly closed in time.

Maybe something like this:

https://github.com/r-lib/testthat/blob/f1dd05cbae2432c1ad8a3db53849ce01819f4172/tests/testthat/test-expect-known.R#L13-L20

`session_info()`:

```
─ Session info ───────────────────────────────────────────────────────────────────────────
 setting  value                       
 version  R version 4.0.3 (2020-10-10)
 os       macOS Big Sur 10.16         
 system   x86_64, darwin17.0          
 ui       RStudio                     
 language (EN)                        
 collate  en_US.UTF-8                 
 ctype    en_US.UTF-8                 
 tz       Europe/Berlin               
 date     2021-01-19                  

[...]

PR #1190, opened 2021-01-20
only use safe on.exit

maxheld83:
As recommended by [testthat](https://testthat.r-lib.org/articles/test-fixtures.html), this PR adds extra arguments to  all `on.exit()`s .

> You should always call it with add = TRUE and after = FALSE. 
> These ensure that the call is added to the list of deferred tasks (instead of replaces) and is added to the front of the stack (not the back, so that cleanup occurs in reverse order to setup). 
> These arguments only matter if you’re using multiple on.exit() calls, but it’s a good habit to always use them to avoid potential problems down the road.

I tried this as a potential fix for #1189 (it didn't help), but maybe it'll still be worthwhile for peace of mind.

Downside: this adds verbosity and may often be unnecessary.

A better solution of test fixtures would probably use `withr::defer()` and existing `withr::with_*()`.
But withr isn't in the dependencies for roxygen2, so maybe this is a medium term improvement.

I think this also only works in >= R 3.5 (the `after` argument is new, I think).
Not sure what the backwards compatibility goal for roxygen2 is.

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -80,7 +80,7 @@ block_create <- function(tokens, call, srcref) {
   if (is.null(pkgenv)) pkgenv <- baseenv()
   evalenv <- new.env(parent = pkgenv)
   roxy_meta_set("evalenv", evalenv)
-  on.exit(roxy_meta_set("evalenv", NULL), add = TRUE)
+  on.exit(roxy_meta_set("evalenv", NULL), add = TRUE, after = FALSE)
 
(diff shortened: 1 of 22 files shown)


hadley:
We need all packages to work on 3.3 and later so a better solution would be to add withr to imports/suggests. But this is something we might do when we switch to testthat 3e. I'll make a note

PR #1194, opened 2021-01-27
Fix suggestion for overriding the default load strategy.

mcol:
The man page for `?load` states that one could add `RoxygenNote: list(load = "source")` to select a different load strategy, but this won't work, as it should say `Roxygen` instead.


hadley:
Thanks!

PR #1213, opened 2021-03-21
Update rd.Rmd

### LINKED ISSUE #1090: @import continuation lines
If we have a package with this single R file

```
#' @import zoo

#' Different name for calling zoo.
#'
#' @params ... passed to zoo.
#' @return zoo object.
#'
#' @export 
zoo2 <- function(...) zoo(...)
```

then running document() on it gives this message
```
zoo2.Rd is missing name/title. Skipping 
```

If I put the @import at the bottom of the file  or put some dummy code line after the @import line then it works as expected not producing that warning. It seems it is regarding the line beginning with Different as being continued from the @import line.

I didn't notice any documentation that would lead one to understand that that is how it works.  It was quite unexpected and took some time to realize what was going on.

monkeywithacupcake:
Add explanation of space and NULL with examples to close #1090 

Not sure if putting this explanation above the list of tags makes sense? Is there a better structure for this section of the docs?

### DIFF
--- a/vignettes/rd.Rmd
+++ b/vignettes/rd.Rmd
@@ -17,6 +17,31 @@ A roxygen __block__ is a sequence of lines starting with `#'` (optionally preced
 
 Text within roxygen blocks can be formatted using markdown or `Rd` commands; see `vignette("rd-formatting")` for details.
 
+Blocks are assumed to continue until a line is `NULL` or no more tags are expected. That is, both `#'` and ` ` on a line will show blank space but continue the block. For example, this code will run:
+```{r}
+#' @import zoo
+NULL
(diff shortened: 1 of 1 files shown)


lastthyme (approved):

monkeywithacupcake:
Can I please get a code review on this? I see that it fails one of the checks, but I 'm not sure why.

hadley:
@monkeywithacupcake thanks for working on this! I took a slightly broader approach — I wanted to keep the intro short, so I just added some info about when a block ends, and then I moved your example into the NAMESPACE vignette (which is where most people will encounter this problem). While I was there I noticed that some of the advice was rather out of date, so I also updated it.

PR #1229, opened 2021-05-13
Match div class to pandoc

hadley:
This makes pkgdown/downlit code simpler since we can assume a more consistent structure across different ways that code blocks are generated.


gaborcsardi (approved):
Cool.

hadley:
cc @maelle

PR #1238, opened 2021-07-08
Clarify export for S3 Methods

MilesMcBain:
I think this is what you mean. At least it seems that when roxygen correctly identifies an S3 method it does not `export()` it in the NAMESPACE file, it only uses the `S3method()` call/directive.

I guess 'export' is overloaded. There's export within the roxygen context, but also within the NAMESPACE file. These things mostly align but sometimes don't. See also: https://github.com/r-lib/roxygen2/issues/1173


hadley:
I meant `export()` in the namespace sense here — you should export methods even if you don't export the generic. Does that make sense?

PR #1260, opened 2021-10-14
Escape curly braces in links

### LINKED ISSUE #1259: Struggling with `{{` links
I'd like to hyperlink mentions of the `{{` operator in the rlang doc because it is not easily searchable.

Here is a counter example with `!!` that works well:

```r
gen <- function(body) {
  header <- "
#' Foo
#' @md
#' @name foo
#' @description
#'"
  text <- paste0(c(header, body, "NULL"), collapse = "\n")
  roxygen2::roc_proc_text(roxygen2::rd_roclet(), text)
}

gen("#' [`!!`][op-inject]")
#> $foo.Rd
#> \name{foo}
#> \alias{foo}
#> \title{Foo}
#> \description{
#> \code{\link[=op-inject]{!!}}
#> }
```

I would like to generate a similar link for `{{`, properly escaped:

```r
#> \code{\link[=op-inject]{\{\{}}
```

Ideally `` [`{{`][foo] `` would work but here is what we get:

```r
gen("#' [`{{`][op-embrace]")
#> Warning: [<text>:5] @description mismatched braces or quotes
#> $foo.Rd
#> \name{foo}
#> \alias{foo}
#> \title{Foo}
#> \description{
#>
#> }
```

Escaping `{` does not work (same result with `\\\\{\\\\}`):

```r
gen("#' [`\\{\\{`][op-embrace]")
#> Warning: [<text>:5] @description mismatched braces or quotes
#> $foo.Rd
#> \name{foo}
#> \alias{foo}
#> \title{Foo}
#> \description{
#>
#> }
```

And unfortunately, trying to insert Rd manually does not work either

```r
gen("#' \\code{\\link[=foo]{\\{\\{}}}}")
#> Warning: [<text>:5] @description mismatched braces or quotes
#> $foo.Rd
#> \name{foo}
#> \alias{foo}
#> \title{Foo}
#> \description{
#>
#> }

gen("#' \\code{\\link[=foo]{\\\\{\\\\{}}}}")
#> $foo.Rd
#> \name{foo}
#> \alias{foo}
#> \title{Foo}
#> \description{
#> \code{\link[=foo]{\\{\\{}}}}
#> }
```

Fixing `` [`{{`][foo] `` so that it works as expected would be great. But I'd appreciate pointers to a workaround as well.

lionel-:
Closes #1259.

First commit escapes the whole link. Second commit only escapes the text nodes of the links. This seems better even though having code and verb nodes in a link doesn't seem to be valid Rd.


lionel-:
> Second commit only escapes the text nodes of the links. 

I guess this only works because ``[`foo`][x]`` inverts the `code` and `link` nodes to produce valid Rd. And then we get a text node even though it's in reality a `code` node.

This is rather convoluted and non-obvious, I'm not sure about that second commit anymore. Maybe we should just find a way to signal the correct node kind for the inverted node (code) to `mdxml_node_to_rd()`. Then we'll get escaping via `mdxml_code()`.

lionel-:
ok last commit should be better.

PR #1282, opened 2022-01-25
added missing body of str_length() in README example

flrd:


hadley:
The point here is to illustrate the roxygen2 style, not the actual implementation of `str_length()`.

PR #1304, opened 2022-03-16
Adapt `github_document()` format for `@includeRmd`

cderv:
This PR is in relation to https://github.com/rstudio/rmarkdown/issues/2331 

I believe the issue could be from the fact that `github_document()` is unnecessary  rendering a HTML self_contained document. `html_preview = FALSE` should be set in the case of `@includeRmd` usage

There may be other adjustment needed so I am creating this as a draft for now until this is tested and sorted out.

### DIFF
--- a/R/rd-include-rmd.R
+++ b/R/rd-include-rmd.R
@@ -45,7 +45,7 @@ roxy_tag_rd.roxy_tag_includeRmd <- function(x, base_path, env) {
 
   rmarkdown::render(
     rmd_path,
-    output_format = rmarkdown::github_document(),
+    output_format = rmarkdown::github_document(html_preview = FALSE),
     output_file = md_path,
     quiet = TRUE,
     envir = new_environment(parent = global_env())


cderv:
Looking at what is produced in the `Rd` file we get something like this 
```
(\figure{https://latex.codecogs.com/png.image?%5Cdpi%7B110%7D&space;%5Cbg_white&space;%5Ceta}{\eta})
````
This does not render well 
image

Are image from the web possible to include in R HTML help page ? 

The solution to this may also be to deactivate completely the math support for **roxygen2** `@includeRmd`

cderv:
If we want to deactivate with `github_document(math_method = NULL)` this can only be done for **rmarkdown** >= 2.12
I have made the addition of `math_method = NULL` conditional to package version. 

Obviously if we decide to make `math_method = NULL` the default in **rmarkdown** then we won't need this.

cderv:
Looking closer at how to do math in Rd document, I believe this should be a better way: https://docs.ropensci.org/katex/reference/math_to_rd.html

So I think it ok to deactivate webtex in the case of this PR.

cderv:
@topepo @EmilHvitfeldt this should solve your issue and also undesired change in Rd file when running `devtools::document()`

@gaborcsardi this PR does not deal with the fact that I don't think a user can adapt the rendering by setting a YAML header in the Rmd included. If we want to support that, this requires other adjustment. 

Not sure who should merge this, but this is ready on roxygen2 side.

gaborcsardi:
Looks good to me.

PR #1311, opened 2022-03-31
Improve feedback from `@inherits` and friends

### LINKED ISSUE #1135: Request: make missing link warnings more verbose
Hi, I'm seeing this warning now pop up in roxygen2 for one of my packages:

```
Warning: Link to unknown topic in inherited text: methods::StructureClasses
```

I'm having trouble locating the source of this warning. Is there a way we can tweak roxygen2 to return which Rd file and line in particular is causing the issue?

hadley:
And considerably reduce the scope of link tweaking.

Fixes #1135.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,6 +21,7 @@ Depends:
     R (>= 3.3)
 Imports: 
     brew,
+    cli,
     commonmark,
     desc (>= 1.2.0),
     digest,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # roxygen2 (development version)
 
+* `@inherit` and friends perform less aggressive link tweaking, eliminating
+  many spurious warnings. Additionally, when you do get a warning, you'll 
+  now always learn which topic it's coming from (#1135).
+
 * Add support for inheriting 'note' fields via `@inherit pkg::fun note` (@pat-s, #1218)
 
 * Problems with the first tag in each block are now reported with the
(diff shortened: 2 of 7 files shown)


gaborcsardi:
I will take a better look tomorrow to be sure.

gaborcsardi (approved):
Looks great!

gaborcsardi:
@hadley Re `find_topic_filename()` (and friends), I tried to make it a no-op:
```r
find_topic_filename <- function(pkg, topic, tag = NULL) {
  return(topic)
}
```
and this doesn't really break anything, except that we need to update/remote 2-3 test cases. I checked on a two days ol R-devel as well.

OTOH there are a couple of things to consider. 

1. It does change the generated manual of course. E.g. for ggplot2:
```
❯ git diff man/ | wc -l
     437
```
so this will be a bit disrupting. 

2. There is a slight, but non-zero chance that a link breaks because the escaping of special characters in topic names works out differently if we don't rewrite the links. This is just a theoretical possibility to keep in mind if we see broken links, so definitely not a major issue.

3. Currently it is nice to get a warning from roxygen if you try to link to a non-existent topic. If we remove the link lookup, then this goes away, and you only learn about broken links from `R CMD check` (unless you turn off the rd xrefs check of course, which you might). So even if we don't rewrite the links, it might make sense to still do the link lookup, and warn for broken links.

hadley:
Yeah, so maybe just easier to leave as is for now.

PR #1313, opened 2022-03-31
Preserve NAMESPACE during pre-processing pass

### LINKED ISSUE #1254: Failures in a file included with `@includeRmd` will nuke the NAMESPACE
If you include a file with `@includeRmd` that fails during `roxygenise()`, then your NAMESPACE will be (mostly) nuked, which often puts you in a state where you can't even `roxygenise()` again to fix it. You end up having to revert the changes made to the NAMESPACE manually, fix the Rmd (or whatever made it fail), and then redocument. (This has bitten the tidymodels team, and me in vctrs quite a few times)

I have a somewhat minimal example here, where this commit changed an Rmd to make it fail, I redocumented, and now the exported function `fn()` is no longer part of NAMESPACE
https://github.com/DavisVaughan/testfailingroxygen2/commit/ad6f49fd34073758c9d69fb88089673051f920e1

The way this occurs is:
- [`roclet_preprocess()`](https://github.com/r-lib/roxygen2/blob/6c1e42f061e9435a14bd3a0e01c7f08fc7aa9160/R/roxygenize.R#L73) has a namespace method that inserts partial information into the NAMESPACE file. It seems to insert information about imports only.
- [`roclet_process()`](https://github.com/r-lib/roxygen2/blob/6c1e42f061e9435a14bd3a0e01c7f08fc7aa9160/R/roxygenize.R#L80) will then run, and the Rd method for that is where the Rmd is rendered and subsequently fails.
- [`roclet_output()`](https://github.com/r-lib/roxygen2/blob/6c1e42f061e9435a14bd3a0e01c7f08fc7aa9160/R/roxygenize.R#L88) never gets to run, which is where the NAMESPACE is finalized, so we end up with the broken partial NAMESPACE that `roclet_preprocess()` wrote.

I see two potential solutions:
- `roxy_tag_rd.roxy_tag_includeRmd()` could `tryCatch()` the `rmarkdown::render()` call? Or maybe pass something to `render()` to make it not fail.
- More generally, I'm not entirely sure why `roclet_preprocess.roclet_namespace()` exists. It seems like the process and output methods for roclet_namespace will basically do the same thing, just with `import_only = FALSE`. If the preprocess method wasn't run, it seems like we wouldn't get into that problematic state from above because we would never write a partial NAMESPACE to disk. I'm probably missing something though https://github.com/r-lib/roxygen2/blob/6c1e42f061e9435a14bd3a0e01c7f08fc7aa9160/R/namespace.R#L34-L63.

hadley:
Now preserves all non-import directives.

Fixes #1254

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,8 @@ Suggests:
     R.methodsS3,
     R.oo,
     rmarkdown,
-    testthat (>= 3.1.2)
+    testthat (>= 3.1.2),
+    withr
 LinkingTo: 
     cpp11
 VignetteBuilder: 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # roxygen2 (development version)
 
+* The NAMESPACE roclet now preserves all existing non-import directives during
+  it's first pre-processing pass. This eliminates the "NAMESPACE has changed"
+  messages and reduces the incidence of namespace borking (#1254).
+
 * Add support for inheriting 'note' fields via `@inherit pkg::fun note` (@pat-s, #1218)
 
 * Problems with the first tag in each block are now reported with the
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -39,8 +39,8 @@ roclet_preprocess.roclet_namespace <- function(x, blocks, base_path) {
     return(x)
   }
 
-  lines <- sort_c(c(lines, namespace_exports(NAMESPACE)))
-  results <- c(made_by("#"), lines)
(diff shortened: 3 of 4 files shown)


gaborcsardi (approved):

hadley:
Maybe I can just unique the lines?

gaborcsardi:
OK, sounds good. It is also OK to keep it as is. I am not too worried about the duplicated expressions, was just wondering if there could be something similar. But even if there is, it is still better than the borked `NAMESPACE`....

PR #1316, opened 2022-04-01
Inherit group params exactly

### LINKED ISSUE #950: Keep argument descriptions together when inheriting
When several arguments are documented together (`@param x,y ...`) and this argument is inherited in another markdown chunk with `@inheritParams` the arguments are split up instead of kept together. Keeping them together in instances where it is possible would make param inheritance more useful

hadley:
Fixes #950

Example with ggplot2 at https://github.com/tidyverse/ggplot2/pull/4791

### DIFF
--- a/R/rd-inherit.R
+++ b/R/rd-inherit.R
@@ -175,7 +175,6 @@ inherit_dot_params <- function(topic, topics, env) {
   # Then pull out the ones we need
   docs <- lapply(inheritors$source, find_params, topics = topics)
   arg_matches <- function(args, docs) {
-    doc_args <- str_split(names(docs), ", ?")
     match <- map_lgl(docs, function(x) x$name %in% args)
     matched <- docs[match]
     setNames(
--- a/R/rd-inherit.R
+++ b/R/rd-inherit.R
@@ -141,24 +141,26 @@ inherit_params <- function(topic, topics) {
   }
 
   for (inheritor in inheritors) {
-    inherited <- find_params(inheritor, topics)
+    inherited_params <- find_params(inheritor, topics)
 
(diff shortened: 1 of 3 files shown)


R/rd-inherit.R
```diff
@@ -174,8 +176,12 @@ inherit_dot_params <- function(topic, topics, env) {
   docs <- lapply(inheritors$source, find_params, topics = topics)
   arg_matches <- function(args, docs) {
     doc_args <- str_split(names(docs), ", ?")
```
DavisVaughan, line 178: Do you need `doc_args` anymore?

tests/testthat/test-rd-inherit.R
```diff
@@ -325,17 +325,35 @@ test_that("multiple @inheritParam inherits from existing topics", {
...
-  expect_equal(length(params), 2)
-  expect_equal(sort(names(params)), c("x", "y"))
+    #' A
+    #' @param x,y X and Y
+    A <- function(x, y) {}
+
+    #' B
+    #'
+    #' @inheritParams A
+    B <- function(x, y) {}"
```
DavisVaughan, line 337: Should you add a test where the args here are `.x` and `.y` to ensure that works? I tested it and it does work right, but a test may keep it from breaking in the future

DavisVaughan (approved):
I don't understand all the technical bits, but it overall looks great! I ran this on iv too and it worked well https://github.com/DavisVaughan/iv/pull/12

tests/testthat/test-rd-inherit.R
hadley: Good catches, thanks

PR #1317, opened 2022-04-05
Use new warn_roxy_tag() and warn_roxy_block()

hadley:
These new functions are powered by cli, given considerably better displays.

A few things to note:

* `markdown()` now checks for incomplete braces; I _think_ this is ok. Certainly it seems inconsistent to only check it for sectioned md, which is what we were doing before.
* I overhauled the generated tags so that that include the file and line from the block. I think this is a nice (if minor) improvement; I did it mainly in order to simplify the warning code in `try_find_topic_in_package()`.
* Converting the tests to use snapshots revealed that a lot of the warning messages were poorly worded and/or ungrammatical.
* Use `warn(parent = err)` makes it so much easier to forward on the underlying problem.

To do:

* [ ] Add news bullet
* [ ] Add RStudio hyperlinks


R/rd-raw.R
```diff
@@ -16,19 +16,21 @@ format.rd_section_rawRd <- function(x, ...) {
   paste(x$value, collapse = "\n")
 }
 
-roxy_tag_eval <- function(tag, env) {
+roxy_tag_eval <- function(tag, env = new.env(emptyenv())) {
```
gaborcsardi, line 19: Do you want
```r
new.env(parent = emptyenv())
```
here?

R/markdown-link.R
```diff
@@ -105,9 +105,10 @@ parse_link <- function(destination, contents, state) {
   if (!all(xml_name(contents) %in% c("text", "softbreak", "linebreak"))) {
     incorrect <- setdiff(unique(xml_name(contents)), c("text", "softbreak", "linebreak"))
 
-    roxy_tag_warning(state$tag,
-      "Links must contain plain text. Problematic link: ", destination
-    )
+    warn_roxy_tag(state$tag, c(
+      "markdown: links must contain plain text",
+      "Problematic link: {destination}"
```
gaborcsardi, line 110: Should this be an `i = `?

R/markdown.R
```diff
@@ -183,7 +183,10 @@ mdxml_children_to_rd <- function(xml, state) {
 mdxml_node_to_rd <- function(xml, state) {
   if (!inherits(xml, "xml_node") ||
       ! xml_type(xml) %in% c("text", "element")) {
-    roxy_tag_warning(state$tag, "Internal markdown translation failure")
+    warn_roxy_tag(state$tag, c(
+      "markdown translation failed",
+      "internal error"
```
gaborcsardi, line 188: Not related to this PR, but I wonder if we could be more helpful here. Even if we just say "please open an issue"?

R/rd-r6.R
```diff
@@ -29,7 +29,9 @@ topic_add_r6_methods <- function(rd, block, env) {
   methods <- add_default_methods(methods)
 
   nodoc <- map_int(methods$tags, length) == 0
-  r6_warning(block, "undocumented R6 method[s]: %s", methods$name[nodoc])
+  if (any(nodoc)) {
+    warn_roxy_block(block, "Undocumented R6 method[s]: {methods$name[nodoc]}")
```
gaborcsardi, line 33: This can be `method{?s}` I believe.

R/rd-r6.R
```diff
@@ -116,15 +123,21 @@ r6_fields <- function(block, r6data) {
 
   # Check for missing fields
   miss <- setdiff(fields, docd)
-  r6_warning(block, "undocumented R6 field[s]: %s", miss)
+  if (length(miss) > 0) {
+    warn_roxy_block(block, "Undocumented R6 field[s]: {miss}")
```
gaborcsardi, line 127: This is probably better with `field{?s}`.

R/rd-r6.R
```diff
@@ -116,15 +123,21 @@ r6_fields <- function(block, r6data) {
...
+  }
 
   # Check for duplicate fields
   dup <- unique(docd[duplicated(docd)])
-  r6_warning(block, "R6 field[s] documented multiple times: %s", dup)
+  if (length(dup) > 0) {
+    warn_roxy_block(block, "R6 field[s] documented multiple times: {dup}")
```
gaborcsardi, line 133: `field{?s}`

R/rd-r6.R
```diff
@@ -116,15 +123,21 @@ r6_fields <- function(block, r6data) {
...
+  }
 
   # Check for extra fields
   xtra <- setdiff(docd, fields)
-  r6_warning(block, "unknown R6 field[s]: %s", xtra)
+  if (length(xtra) > 0) {
+    warn_roxy_block(block, "Unknown R6 field[s]: {xtra}")
```
gaborcsardi, line 139: `field{?s}`

R/rd-r6.R
```diff
@@ -156,11 +164,15 @@ r6_active_bindings <- function(block, r6data) {
 
   # Check for missing bindings
   miss <- setdiff(active, docd)
-  r6_warning(block, "undocumented R6 active binding[s]: %s", miss)
+  if (length(miss) > 0) {
+    warn_roxy_block(block, "Undocumented R6 active binding[s]: {miss}")
```
gaborcsardi, line 173: `binding{?s}`

R/rd-r6.R
```diff
@@ -156,11 +164,15 @@ r6_active_bindings <- function(block, r6data) {
...
+  }
 
   # Check for duplicate bindings
   dup <- unique(docd[duplicated(docd)])
-  r6_warning(block, "R6 active binding[s] documented multiple times: %s", dup)
+  if (length(dup) > 0) {
+    warn_roxy_block(block, "R6 active binding[s] documented multiple times: {dup}")
```
gaborcsardi, line 179: `binding{?s}`

tests/testthat/test-rd-r6.R
```diff
@@ -331,19 +328,13 @@ test_that("integration test", {
...
+  expect_snapshot(
+    res <- roclet_process(roc, blocks = blocks, env = env, base_path = test_path())
   )
 
-  # Warnings
-  expect_snapshot_output(sort(roxy_warnings))
-
+  # # Warnings
+  # expect_snapshot_output(sort(roxy_warnings))
+  #
```
gaborcsardi, line 337: Why keep this in comments?

gaborcsardi (approved):

gaborcsardi:
Looks great! Not only are the warnings better, the code that generates them is simpler as well. Left some comments, mostly cosmetics.

R/markdown.R
```diff
@@ -221,11 +224,17 @@ mdxml_node_to_rd <- function(xml, state) {
 }
 
 mdxml_unknown <- function(xml, tag) {
-  roxy_tag_warning(tag, "Unknown xml node: ", xml_name(xml))
+  warn_roxy_tag(tag, c(
+    "markdown translation failed",
+    x = "Internal error: unknown xml node {xml_name(xml)}"
```
gaborcsardi, line 229: We could be more helpful here as well, maybe suggest opening an issue?

R/rd-r6.R
hadley: Doesn't that require the object/length to before the plural suffix?

R/tag.R
```diff
@@ -89,5 +105,26 @@ print.roxy_tag <- function(x, ...) {
...
+#' @export
+#' @rdname roxy_tag
+warn_roxy_tag <- function(tag, message, ...) {
+  message[[1]] <- paste0(
+    "[", tag$file, ":", tag$line, "] ",
+    "@", tag$tag, " ",
+    if (is.null(tag$raw)) ("(automatically generated) "),
+    message[[1]]
+  )
+  cli::cli_warn(message, ..., .envir = parent.frame())
```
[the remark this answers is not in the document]
hadley: I tried it and it wasn't much improvement and then I couldn't figure out how to nicely add it back into `message` without having the first element double processed.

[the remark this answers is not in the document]
gaborcsardi: Oh, right, yeah, that is hard.

PR #1319, opened 2022-04-06
Fixes for HTML 5

hadley:


R/markdown.R
```diff
@@ -256,6 +256,7 @@ mdxml_code_block <- function(xml, state) {
   info <- xml_attr(xml, "info")[1]
   if (is.na(info) || nchar(info[1]) == 0) info <- NA_character_
   paste0(
+    "\n\n",
```
gaborcsardi, line 259: Doesn't this add an extra empty line? AFAIR in Rd you should not have an empty line or even a line break before `\preformatted{}` because that adds an empty line.

Why is it needed, btw?

R/rd-r6.R
```diff
@@ -249,38 +250,38 @@ r6_inherited_method_list <- function(block, r6data) {
...
   )
 }
 
 r6_method_begin <- function(block, method) {
   nm <- r6_show_name(method$name)
   c(
     "\\if{html}{\\out{<hr>}}",
-    paste0("\\if{html}{\\out{<a id=\"method-", nm, "\"></a>}}"),
-    paste0("\\if{latex}{\\out{\\hypertarget{method-", nm, "}{}}}"),
+    paste0("\\if{html}{\\out{<a id=\"method-", method$class, "-", nm, "\"></a>}}"),
```
gaborcsardi, line 283: To avoid breaking cross package links, can we define both anchors here?

R/rd-r6.R
```diff
@@ -249,38 +250,38 @@ r6_inherited_method_list <- function(block, r6data) {
     "><summary>Inherited methods</summary>"
   )
 
-  c("\\if{html}{",
-    paste0("\\out{", details, "}"),
-    "\\itemize{",
+  c("\\if{html}{\\out{",
+    "<ul>",
```
gaborcsardi, line 254: By writing this in HTML, we rely on how the current Rd -> HTML translation works. If they change it from `<ul>` to something else, or they add classes, etc. then our generated docs will be different.

gaborcsardi, line 254: Seems like `details` is not emitted any more?

tests/testthat/roxygen-block-3-B.Rd
```diff
@@ -36,22 +36,21 @@ Class B details.
...
-}
-\out{</details>}
-}
+\item \href{#method-B-meth1}{\code{B$meth1()}}
+\item \href{#method-B-meth4}{\code{B$meth4()}}
+\item \href{#method-B-clone}{\code{B$clone()}}
+}
+}
+\if{html}{\out{
+<ul>
```
gaborcsardi, line 45: The opening `<details>` tag is missing from here, see above.

gaborcsardi (commented):
Good to see that this did not need a lot of changed. I left some comments.

R/markdown.R
hadley: This makes `\preformatted` start its own paragraph (which doesn't get wrapped in `<p>`), otherwise you it starts before the text paragraph ends and you get `<pre></p>` in the rendered HTML.

R/rd-r6.R
hadley: Oh hmmmm, we can't do that because it'll lead to the same HTML 5 error.

gaborcsardi: Oh, right, of course.

PR #1324, opened 2022-04-08
Always generate div.sourceCode

### LINKED ISSUE #1234: Add sourcecode to code blocks even when there is no info?
I.e. a block like

````
```
bla
```
````

would get `<div class="sourceCode">`

hadley:
Fixes #1234

@gaborcsardi do you see any reason not to do this?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # roxygen2 (development version)
 
+* Markdown code blocks are always wrapped in `<div class="sourceCode">`
+  even if the language is unknown (#1234).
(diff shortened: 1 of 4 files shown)


gaborcsardi:
Yeah, seems harmless. But then what's up with the CI failures?

What does the `sourceCode` class buy us, btw? Something in pkgdown?

hadley:
@gaborcsardi yeah, this just makes things a little more consistent for pkgdown. The failures were a dumb mistake on my end.

PR #1329, opened 2022-04-09
Convert primary user feedback to cli

hadley:
These are the major bits of feedback generated by roxygen2. I'll do another pass on the smaller bits afterwards, but this already required a decent amount of refactoring so it's at a good place to review.


R/roxygenize-setup.R
```diff
@@ -0,0 +1,73 @@
...
+      cli::cli_rule()
+      cli::cli_inform(c(
+        "Changes in roxygen2 7.0.0:",
+        "* `%` is now escaped automatically in Markdown mode.",
+        "Please carefully check .Rd files for changes"
+      ))
+      cli::cli_rule()
+    }
+
+    cli::cli_inform("Setting {.var RoxygenNote} to {cur}")
```
gaborcsardi, line 49: FWIW this was `{.str {cur}}` above.

R/roxygenize.R
```diff
@@ -30,16 +30,10 @@ roxygenize <- function(package.dir = ".",
                        load_code = NULL,
                        clean = FALSE) {
 
-  base_path <- normalizePath(package.dir, mustWork = TRUE)
```
gaborcsardi, line 33: I wonder if we should still call `normalizePath()`?

gaborcsardi (approved):
Looks good. I had two small comments.

R/roxygenize.R
hadley: What are the advantages? Keeping a relative path makes the paths in feedback much more useful.

gaborcsardi: At least on Windows `normalizePath()` does various conversions, so if we don't call it now, then the behaviour probably changes in some cases.

OTOH if it is only used for reporting then that is fine.

hadley: Hmmm, if I re-add it, none of the tests fail, so I must've only needed it in an interim state. Seems better to keep to minimise unintended consequences.

PR #1330, opened 2022-04-11
Finish remainder of cli conversion

### LINKED ISSUE #1154: Use cli for feedback to user

### LINKED ISSUE #1207: `write_if_different()` prints rather than a message
Is it intentional that `write_if_different()` uses `cat()` instead of `inform()` or some other type of message? I noticed that `roxygenise()` requires both `capture.output()` and `suppressMessages()` to shush.

hadley:
Fixes #1154. Fixes #1207.

### DIFF
--- a/R/load.R
+++ b/R/load.R
@@ -82,11 +82,11 @@ find_load_strategy <- function(x, option = roxy_meta_get("load", "pkgload")) {
   if (is.null(x)) {
     x <- option
     if (!is.character(x) || length(x) != 1) {
-      abort("roxygen2 `load` option must be a string")
+      cli::cli_abort("roxygen2 `load` option must be a string")
     }
   } else {
     if (!is.character(x) || length(x) != 1) {
-      abort("`load_code` must be a string or function")
+      cli::cli_abort("`load_code` must be a string or function")
     }
   }
 
@@ -94,6 +94,6 @@ find_load_strategy <- function(x, option = roxy_meta_get("load", "pkgload")) {
     pkgload = load_pkgload,
     source = load_source,
     installed = load_installed,
-    abort("Unknown value of `load` option")
+    cli::cli_abort("Unknown value of `load` option")
   )
 }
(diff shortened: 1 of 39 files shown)


R/load.R
```diff
@@ -82,18 +82,18 @@ find_load_strategy <- function(x, option = roxy_meta_get("load", "pkgload")) {
   if (is.null(x)) {
     x <- option
     if (!is.character(x) || length(x) != 1) {
-      abort("roxygen2 `load` option must be a string")
+      cli::cli_abort("roxygen2 `load` option must be a string")
```
gaborcsardi, line 85: To we want to write 
```
      cli::cli_abort("roxygen2 {.code load} option must be a string")
```
here and also below?  At some point we should write up some conventions.

R/rd-template.R
```diff
@@ -42,7 +42,9 @@ template_find <- function(base_path, template_name) {
   path_exists <- file.exists(path)
 
   if (!any(path_exists)) {
-    stop("Can't find template '", template_name, "'", call. = FALSE)
+    # This should really use warn_roxy_tag() but it's not worth refactoring
+    # for this rarely used feature
+    cli::cli_abort("Can't find template {.str {template_name}}'", call = NULL)
```
gaborcsardi, line 47: Spurious single quote at the end?

gaborcsardi (approved):
Two small comments.

gaborcsardi:
I have no clue what's up with the windows CI. It is unlikely that both windows jobs have a corrupt cache.

R/load.R
hadley: Or `.arg`?

hadley:
@gaborcsardi hmmm, I polished that test and I guess I must've messed it up?

hadley:
@gaborcsardi ooops yeah, I broke the skip — the windows GitHub runner must have `autocrlf = true` which the skip was supposed to avoid.

PR #1333, opened 2022-04-15
remove duplicated authors

DanChaltiel:
As requested in https://github.com/r-lib/roxygen2/issues/1269, here is a PR to remove duplicated authors.

Learned a lot about R6 objects in the process, this was quite fun 😀 

May I also suggest tweaking `format_collapse()` so that you can change the collapsing string? I think a single `\n` would be nicer in `format.rd_section_author()`. (I can make another PR for this one)


R/rd-author.R
```diff
@@ -0,0 +1,7 @@
+
+#' @export
```
hadley, line 2: Given how simple this is, I think you could could keep it in `rd-markdown.R`

hadley (commented):
Thanks for working on this!

Could you please also re-document (to update `NAMESPACE`) and add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

R/rd-author.R
DanChaltiel: I agree this is quite simple, but `rd-markdown.R` has such homogenous content I felt bad to blemish it with this one. 
Besides, all `merge.xxx()` functions belong in a specific `rd-xxx.R` file.
Are you sure you want me to put it in `rd-markdown.R`?

hadley: Yes 😄

hadley:
Thanks!

PR #1351, opened 2022-05-24
Make evaluation more reproducible

### LINKED ISSUE #1349: Ensure reproducible output when running Rmd chunks
e.g. copying the key bits from `local_reproducible_output()` to suppress colours.

Will need fixes to `roxy_tag_rd.roxy_tag_includeRmd`, `eval_code_nodes`, `roxy_tag_eval`.

hadley:
And suppress various cli capabilities that you don't want inside docs

Fixes #1349

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # roxygen2 (development version)
 
+* Code evaluated in inline markdown code chunks and `@eval`/`@evalRd`/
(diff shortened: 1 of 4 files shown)


gaborcsardi (approved):
I think this is good. OTOH at some point it would be nice to have a way to include the better output in the HTML, and fall back to the simple output for PDF and text output.

PR #1356, opened 2022-06-01
Use .empty() to test empty strings

MichaelChirico:
See https://clang.llvm.org/extra/clang-tidy/checks/readability-container-size-empty.html


hadley:
Thanks!

PR #1363, opened 2022-06-16
Add missing span closing tag

hadley:


gaborcsardi (approved):
I wonder how we did not notice this before...

PR #1383, opened 2022-07-07
Only use nbsp for old style usage

### LINKED ISSUE #1342: CRAN Warning when Non-breaking space in default function argument value
Hi,

Thanks for your time and work on the roxygen2 package. It makes developing R packages much more convenient to me.

I have a function that takes a string with a default value that contains non-breaking spaces. When I use `devtools::document()` and afterwards `devtools::check()` I get a WARNING pointing out a mismatch between my code and the help page.

A minimal function with roxygen documentation that gives the WARNING:

```r
#' A dummy function
#'
#' @param str_with_nbsp A string, potentially with non-breaking spaces
#'
#' @return Not relevant
#' @export
#'
myfun <- function(str_with_nbsp = "x\u00a0=\u00a04") {print(str_with_nbsp)}
```

You can find a minimal package that includes this function, with github actions set up here: https://github.com/zeehio/testnbsp

The CRAN WARNING (not that helpful to read given that the regular space and the non-breaking space look the same...)

```
   Codoc mismatches from documentation object 'myfun':
   myfun
     Code: function(str_with_nbsp = "x = 4")
     Docs: function(str_with_nbsp = "x = 4")
     Mismatches in argument default values:
       Name: 'str_with_nbsp' Code: "x = 4" Docs: "x = 4"
```

Here is some reproducible example that could be easily converted into a unit test for this bug:

``` r
myfun <- function(str_with_nbsp = "x\u00a0=\u00a04") {print(str_with_nbsp)}
# The wrap_usage function replaces all my nbsp characters by regular characters
fun_usage <- roxygen2:::wrap_usage("myfun", identity, formals = formals(myfun))
fun_usage
#> <rd> myfun(str_with_nbsp = "x = 4")
# The nbsp 00a0 character was replaced by regular spaces:
grepl(pattern = "\u00a0", as.character(fun_usage)) # I expected this to be TRUE
#> [1] FALSE
```

<sup>Created on 2022-05-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

After looking at the code, I have been able to see that the replacement happens here: https://github.com/r-lib/roxygen2/blob/630ab47cff290e94380e0bfa57d6de964c722af3/R/rd-usage.R#L145

It feels like first roxygen2 introduces: `\u{A0}=\u{A0}` (in `args_string`). Afterwards it replaces all nbsp with regular spaces. A simple solution would be change the final search and replace so instead of `out <- gsub("\u{A0}", " ", out, useBytes = TRUE)` it becomes `out <- gsub("\u{A0}=\u{A0}", "=", out, useBytes = TRUE)`, so it would replace the non breaking spaces that roxygen adds, but not the ones that were present before.

However I don't know if `\u{A0}` and other UTF-8 characters are still problematic on some Windows locales or not... maybe with the recent changes in Windows support many of these workarounds can now be simplified...

Any help will be very much appreciated :pray:

hadley:
Fixes #1342. Closes #1343.

@zeehio I took a slightly different approach because the nbsp are only needed for the old style usage wrapping.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # roxygen2 (development version)
 
+* Automated usage no longer mangles nbsp in default arguments (#1342).
+
 * Give useful error if `@describeIn` can't combine docs (#1366).
 
 * Code evaluated in inline markdown code chunks and `@eval`/`@evalRd`/
--- a/R/rd-usage.R
+++ b/R/rd-usage.R
@@ -111,8 +111,8 @@ usage_args <- function(args) {
   map_chr(args, arg_to_text)
 }
 
-args_string <- function(x) {
-  sep <- ifelse(x != "", "\u{A0}=\u{A0}", "")
+args_string <- function(x, space = " ") {
+  sep <- ifelse(x != "", paste0(space, "=", space), "")
   arg_names <- escape(auto_backtick(names(x)))
   paste0(arg_names, sep, escape(x))
 }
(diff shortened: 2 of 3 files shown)


tests/testthat/_snaps/rd-usage.md
```diff
@@ -27,8 +27,8 @@
       x,
       y,
       xy = "abcdef",
-      xyz = c(`word word word word` = "abcdef", `word word word` = "abcdef",
-        `word word word` = "abcdef", `word word word` = "abcdef")
+      xyz = c(`word word word word` = "abcdef", `word word word` = "abcdef", `word word word`
```
hadley, line 30: This only looked good by coincidence before, because the C++ wrapping code counts bytes, not characters, and `\u{A0}` is two bytes.

zeehio:
Great! For what is worth I reviewed the pull request  and it looks good to me 😀🥳

Thank you very much!

PR #1390, opened 2022-07-10
Allow overriding default knitr chunk options

gaborcsardi:
I think this would be useful to avoid repetition. I can try to write some test case.

### DIFF
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -110,7 +110,13 @@ eval_code_node <- function(node, env) {
     # write knitr markup for fenced code
     text <- paste0("```", xml_attr(node, "info"), "\n", xml_text(node), "```\n")
   }
-  roxy_knit(text, env, knitr_chunk_defaults)
+
+  chunk_opts <- modifyList(
+    knitr_chunk_defaults,
+    roxy_meta_get("knitr_chunk_options", NULL)
+  )
+
+  roxy_knit(text, env, chunk_opts)
 }
 
 knitr_chunk_defaults <- list(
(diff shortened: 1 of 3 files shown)


hadley:
What about reading from `knitr::opts_chunk$get()` so we could avoid having to maintain the state ourselves?

What do you think the scope should be for the options? A single `.Rd` or the whole package?

gaborcsardi:
IDK, it just seems simpler to specify the defaults in the config, and then use them in the whole package.

So I think the scope should be the whole package, for simplicity.

hadley:
Yeah, that makes sense. Can you also add a few words to the `rd-formatting.Rmd` vignette in the code chunks section?

gaborcsardi:
@hadley all done, I believe. 

We need to run the test in a subprocess, unfortunately, because it messed up the roxygen pptions we rely on in other test cases. Alternatively we could add a `knitr_chunk_options` option to roxygen2 itself, maybe that's better?

tests/testthat/test-options/man/roxygen/meta.R
```diff
@@ -1,3 +1,6 @@
 list(
-  markdown = TRUE
+  markdown = TRUE,
+  knitr_chunk_options = list(
+    myoption = 1:10
```
hadley, line 4: Maybe also set `comment = "###"` here just to check that these override the chunk defaults?

hadley:
Maybe we should have a `local_roxy_meta_set()`? Then I think you could just `roc_proc_text()` instead of having to set up a separate package?

PR #1399, opened 2022-07-11
Allow restricting image formats

### LINKED ISSUE #1398: Allow conditional or alternative images in markdown
WRE suggests SVG + PDF for nice images in HTML and PDF:
```
\if{html}{\figure{Rlogo.svg}{options: width=100 alt="R logo"}}
\if{latex}{\figure{Rlogo.pdf}{options: width=0.5in}}
```

As a first step we could support conditional images. My use case is adding SVG images to the HTML manual, and I don't mind if they are omitted from the PDF manual.

gaborcsardi:
This is currently opt in. If you set `restrict_image_formats`
to `TRUE`, then SVG images are HTML-only and PDF images are
PDF-only. Other file types are not restricted.

We could try to make it opt out, because SVGs do not work in
LaTeX and PDFs most of the time do not work in HTML.

If it is opt in, then we don't necessarily need a NEWS item
to try this out, but if it ends up being opt out, then we
do and we also need to document it,

Closes #1398.


R/options.R
```diff
@@ -30,6 +30,10 @@
 #'
 #' * `knitr_chunk_options`: default chunk options used for knitr.
 #'
+#' * `restrict_image_formats`: if `TRUE` then PDF images are only included in the
+#'   PDF manual, and SVG images are only included in the HTML manual.
+#'   (This only applies to images generated via markdown.)
```
hadley, line 35:
```suggestion
#'   (This only applies to images supplied via markdown.)
```

tests/testthat/test-markdown.R
```diff
@@ -648,3 +648,19 @@ test_that("can override default options", {
   ")[[1]]
   expect_match(out$get_section("description")$value, "###", fixed = TRUE)
 })
+
+test_that("image formats work", {
+  local_roxy_meta_set("restrict_image_formats", TRUE)
+
```
hadley, line 654:
```suggestion
```

tests/testthat/test-markdown.R
```diff
@@ -648,3 +648,19 @@ test_that("can override default options", {
...
+  expect_snapshot(
+    roc_proc_text(rd_roclet(), "
+      #' Title
+      #'
+      #' ![](example.svg \"Plot title 1\")
+      #' ![](example.pdf \"Plot title 2\")
+      #' ![](example.PNG \"Plot title 3\")
+      #' @md
+      foo <- function() { }
+    ")
```
hadley, line 664:
```suggestion
    ")[[1]]
```

Will need to re-run snapshots.

PR #1429, opened 2022-09-16
use find.package() over system.file() to check installed

MichaelChirico:
More direct &  should be more readable (not clear if the upstream rlang issue is still a blocker)


hadley:
I've taken so long to get to this that I think switching back to rlang will be easier: #1453

PR #1458, opened 2023-01-16
Throw an error for @importFrom unknown export

### LINKED ISSUE #1409: Check imports when generating the NAMESPACE file
Ie. for an `@importFrom pkg fun` we could check if `fun` is an exported function of `pkg`. 

Otherwise R will throw a warning when the package is loaded, and it is sometimes hard to deduct where the warning is coming from.

MichaelChirico:
Closes #1409.

This is a rough / off-the-cuff version. Wanted to checkpoint here to get high-level feedback, there's still TODO:

 - [x] Adapt to internal package style (e.g. `cli`)
 - [x] Add tests
 - [x] Add NEWS


hadley (commented):
This seems reasonable to me.

MichaelChirico:
I can't make snapshots from my current machine, so I'll try and pull them from the GHA artifacts.

Should be ready for review anyway.

R/namespace.R
```diff
@@ -229,6 +229,14 @@ roxy_tag_parse.roxy_tag_importClassesFrom <- function(x) {
 }
 #' @export
 roxy_tag_ns.roxy_tag_importClassesFrom <- function(x, block, env, import_only = FALSE) {
+  pkg <- x$val[1L]
+  if (requireNamespace(pkg, quietly = TRUE)) {
+    importing <- x$val[-1L]
+    unknown_idx <- !importing %in% getNamespaceExports(pkg)
+    if (any(unknown_idx)) {
+      cli::cli_abort("Unknown {cli::qty(sum(unknown_idx))} export{?s} in {.code @importFrom {pkg}}: {.code {importing[unknown_idx]}}")
```
hadley, line 237: Sorry, I remembered that this should use `roxy_warn_tag()` as that will also give information about where the tag was found.

R/namespace.R
```diff
@@ -229,6 +229,14 @@ roxy_tag_parse.roxy_tag_importClassesFrom <- function(x) {
 }
 #' @export
 roxy_tag_ns.roxy_tag_importClassesFrom <- function(x, block, env, import_only = FALSE) {
+  pkg <- x$val[1L]
```
hadley, line 232: I think you meant to do this in `roxy_tag_ns.roxy_tag_importFrom()`

MichaelChirico: :sweat_smile: indeed

R/namespace.R
MichaelChirico: Hmm I had stayed away from that since `warn_roxy_tag()` will still proceed to write the invalid imports in the NAMESPACE. Switched to `warn_roxy_tag()` for now but is that the best behavior?

R/namespace.R
```diff
@@ -238,6 +238,14 @@ roxy_tag_parse.roxy_tag_importFrom <- function(x) {
 }
 #' @export
 roxy_tag_ns.roxy_tag_importFrom <- function(x, block, env, import_only = FALSE) {
+  pkg <- x$val[1L]
+  if (requireNamespace(pkg, quietly = TRUE)) {
+    importing <- x$val[-1L]
+    unknown_idx <- !importing %in% getNamespaceExports(pkg)
+    if (any(unknown_idx)) {
+      warn_roxy_tag(x, "Unknown {cli::qty(sum(unknown_idx))} export{?s} in {.code @importFrom {pkg}}: {.code {importing[unknown_idx]}}")
```
hadley, line 246: You'll also need to modify `x$val` here to avoid generating the bad input.

MichaelChirico: i.e. just drop the unknown exports after throwing the warning?

great idea! makes me much happier with a warning instead of an error

hadley: Yeah, exactly. I'd say that's the general roxgen2 strategy for problematic user input.

hadley:
Thanks!

R/namespace.R
```diff
@@ -238,6 +238,15 @@ roxy_tag_parse.roxy_tag_importFrom <- function(x) {
 }
 #' @export
 roxy_tag_ns.roxy_tag_importFrom <- function(x, block, env, import_only = FALSE) {
+  pkg <- x$val[1L]
+  if (requireNamespace(pkg, quietly = TRUE)) {
+    importing <- x$val[-1L]
+    unknown_idx <- !importing %in% getNamespaceExports(pkg)
```
MichaelChirico, line 244: two issues with this PR in retrospect as surfaced by #1570:

 1. `importing` can be like `"%>%"` or `` `%>%` `` or `'%>%'` but `getNamespaceExports()` will always just be the plain symbol (`%>%`)
 2. `importing[!unknown_idx]` might drop everything --> output becomes `importFrom($PKG,)`

How would you recommend approaching these?

 1. Is a bit tricky because we have to normalize the quotes, I guess we can do `gsub("^(`|'|\")(.*)\\1$", "\\1", importing)`. This could be a helper alongside [`has_quotes()`](https://github.com/r-lib/roxygen2/blob/767ec1b1c4dc43b4245ed4ee2334445968768f17/R/utils.R#L181) Initially, I was worried about collision where you try and import a really weirdly-named object like `as.name("'%>%'")`, but I guess {roxygen2} has always made that difficult. And a quick search doesn't turn up any usage of this anyway: https://github.com/search?q=lang%3AR+%2F%5Cn%5Cs*%28%22%27.*%27%22%7C%27%22.*%22%27%7C%60%22.*%22%60%7C%60%27.*%27%60%7C%22%60.*%60%22%7C%27%60.*%60%27%29%5Cs*%3C-%2F+-path%3A.Rd&type=code
 2. Should this method just `return(NULL)` if `all(unknown_idx)`?

hadley: 1. Yeah, I think `strip_quotes()` seems reasonable.
2. Yes, I think sol.

MichaelChirico: thanks! that's basically how I did #1574 so let's continue there.

PR #1471, opened 2023-03-11
Add alt text to logo in README

ateucher:
This appears to be the only place where alt-text is required. 
As discussed, I set the alt in the logo to `""` since it is purely decorative.

#1467


hadley:
Our convention (at least from the usethis function) now appears to be to use "{package} website".

PR #1481, opened 2023-05-11
remove duplicated text (belonging to output rather than examples)

espinielli:
The removed text belongs to the output section rather than the examples one.


hadley:
I think that actually needs to say that all functions need examples for CRAN submission.

PR #1488, opened 2023-06-12
Ignore self-import directives

lionel-:
If I'm not missing anything that's all that is needed to provide a solution for https://github.com/r-lib/usethis/issues/1853. With this PR, an `#' @import(foo)` directive is ignored when documenting the package foo. This makes it possible for standalone files to define their self dependencies.

### DIFF
--- a/R/roxygenize-setup.R
+++ b/R/roxygenize-setup.R
@@ -25,12 +25,22 @@ roxygen_setup <- function(path = ".",
   man_path <- file.path(path, "man")
   dir.create(man_path, recursive = TRUE, showWarnings = FALSE)
 
-  local_options(
-    "roxygen2:::package" = desc::desc_get("Package", path),
-    .frame = frame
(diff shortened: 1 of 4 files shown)


lionel-:
The snapshot failures also happen with main.

R/roxygenize-setup.R
```diff
@@ -23,6 +25,11 @@ roxygen_setup <- function(path = ".", cur_version = NULL) {
   man_path <- file.path(path, "man")
   dir.create(man_path, recursive = TRUE, showWarnings = FALSE)
 
+  local_options(
+    "roxygen2:::package" = desc::desc_get("Package", path),
```
hadley, line 29: Maybe we should make this an env var so it also solves #1517?

lionel-: Now done

hadley (approved):

PR #1496, opened 2023-07-04
Add examples for @rdname to 'reuse' vignette

### LINKED ISSUE #1495: Documentation feedback: Difference between @rdname and @describeIn unclear in reuse.Rmd
Received this feedback from a user (related to #1493).

Scanning the document, I also don't see a clear distinction made between the two. There are several `@describeIn` examples, but the only `@rdname` examples are being used to highlight the `@order` feature.

IMO a ``### `@rdname` `` subsection paralleling this for `@describeIn` would be appropriate:

https://github.com/r-lib/roxygen2/blob/a34bdf7c7b952068b1f7742b7b34d1eac78e54da/vignettes/reuse.Rmd#L38

At a minimum, linking out to a more complete description of the two tags would help.

MichaelChirico:
Closes #1495.

I am far from an expert here, and only have a vague sense of when to use one vs. the other myself. The "vibe" I have is `@rdname` is better when the functions will be "on equal footing" on the shared page, where as `@describeIn` is more apt for "corollary methods" of a main topic. If that's right, it may be useful to convey that in these sections as well.


hadley:
Thanks for getting me started on this. Can you take a look at my revisions and let me know if they make sense to you?

vignettes/reuse.Rmd
```diff
@@ -35,6 +35,53 @@ You can document multiple functions in the same file by using either `@rdname` o
 It's a technique best used with care: documenting too many functions in one place leads to confusion.
 Use it when all functions have the same (or very similar) arguments.
 
+### `@rdname`
+
+Use `@rdname <destination>` to include multiple functions in the same page.
+The functions will be collected and share the same `@title`, `@description`, `@examples`, etc.
+
+There are two basic ways to use it.
+You can create a dummy documentation block and then add the functions to it.
```
MichaelChirico, line 44: nit: recently have been avoiding the term 'dummy' (e.g. https://blog.leeandlow.com/2012/08/16/when-did-the-word-dummy-become-derogatory/ / https://source.android.com/docs/setup/contribute/respectful-code#term-examples / https://nvlpubs.nist.gov/nistpubs/ir/2021/NIST.IR.8366.pdf), usually 'placeholder' serves the purpose equally well.

vignettes/reuse.Rmd
```diff
@@ -35,6 +35,53 @@ You can document multiple functions in the same file by using either `@rdname` o
...
+
+#' @rdname log
+#' @export
+log2 <- function(x) log(x, 2)
+
+#' @rdname log
+#' @export
+ln <- function(x) log(x, exp(1))
+```
+
```
MichaelChirico, line 84: do we need `#' @export` tags in the next section (L94-103)?

vignettes/reuse.Rmd
```diff
@@ -35,6 +35,53 @@ You can document multiple functions in the same file by using either `@rdname` o
 It's a technique best used with care: documenting too many functions in one place leads to confusion.
 Use it when all functions have the same (or very similar) arguments.
 
+### `@rdname`
+
+Use `@rdname <destination>` to include multiple functions in the same page.
+The functions will be collected and share the same `@title`, `@description`, `@examples`, etc.
```
MichaelChirico, line 41: > The functions will be collected and share the same `@title`, `@description`, `@examples`, etc.

Sentence reads a bit awkwardly to me, I think it's the passive voice? Maybe "roxygen will collect functions marked with the same `<destination>` and give them the same `@title`..."?

vignettes/reuse.Rmd
```diff
@@ -35,6 +35,53 @@ You can document multiple functions in the same file by using either `@rdname` o
...
+#' @export
+cos_ish <- function(x) 1 - x^2 / 2
+
+#' @rdname trig
+#' @export
+tan_ish <- function(x) x + x^3 / 3
+```
+
+Alternatively, you can add docs to an existing function.
+This tends to work better if you have a "primary" function with some variants or some helpers.
```
MichaelChirico, line 67: I'm not sure we're _quite_ there yet, now we have in `@rdname`

> This tends to work better if you have a "primary" function with some variants or some helpers.

Vs in `@describeIn`

> You can document several functions which perform slight variations of a task or which use different defaults

Do we need to distinguish these two "variant"/'variation" use cases further? Or are they \~somewhat\~ interchangeable?

MichaelChirico (commented):
Thanks, looking much better already!

vignettes/reuse.Rmd
hadley: TBH I've mostly forgotten what `@describeIn` does and why you might use it. I think the main difference is that it attempts to build the single description for you by creating a bulleted list. But I'm not sure this is now possible to do in a compelling way, and I think you're generally better off using `@rdname` with a custom `@description`.

MichaelChirico: That seems reasonable enough to me. OTOH, as someone looking at a lot of other peoples' code, it's something I come across fairly often -- it would be nice to point to a canonical reference during code review urging one or the other in each case.

I see 1000s of usages on CRAN:

https://github.com/search?q=lang%3AR+%2F%23%27%5Cs%40describeIn%2F+org%3Acran&type=code

Here's the only file with usage in tidyverse:

https://github.com/tidyverse/reprex/blob/b904f73989583e66cb02fd0199df6cede5be1c3f/R/reprex-undo.R#L22

Vs. 10 packages using it from r-lib:

https://github.com/search?q=lang%3AR+%2F%23%27%5Cs%40describeIn%2F+org%3Ar-lib&type=code

vignettes/reuse.Rmd
hadley: Good point. And I think "standalone" is more clear here anyway.

vignettes/reuse.Rmd
hadley: Ok, I've made it more clear that I don't recommend it, and I'll fix the reprex usage momentarily.

hadley: https://github.com/tidyverse/reprex/pull/452

MichaelChirico:
Thanks!

PR #1499, opened 2023-07-07
Correct `@inheritParams` vs `@inheritDotParams template

olivroy:
If I understand correctly, `@inheritDotParams` doesn't need extra arguments in autocomplete.
This should be `@inheritParams` instead


olivroy:
Sorry, went too fast. Closing

Issue #1485, opened 2023-05-29
typos


musvaage:
Increment
colon
forgotten
generate
insensitive
preferably
section

```
$ grep -nr Incremenet roxygen2
roxygen2/src/parser2.cpp:120:      // Incremenet curRow for non-roxygen comments at start of block
$ grep -nr colom roxygen2
roxygen2/tests/testthat/test-rd-section.R:1:test_that("warn if forgotton colom", {
roxygen2/tests/testthat/_snaps/rd-section.md:1:# warn if forgotton colom
$ grep -nr forgotton roxygen2
roxygen2/tests/testthat/test-rd-section.R:1:test_that("warn if forgotton colom", {
roxygen2/tests/testthat/_snaps/rd-section.md:1:# warn if forgotton colom
$ grep -nr geenrate roxygen2
roxygen2/vignettes/rd-formatting.Rmd:305:This may geenrate markdown not supported by roxygen2.
$ grep -nr insentive roxygen2
roxygen2/tests/testthat/test-rd-template.R:9:  # On case-insentive file systems, will find upper case version first
$ grep -nr preferrably roxygen2
roxygen2/.github/SUPPORT.md:45:    (preferrably github) and link to it in the issue. The more you can do to
$ grep -nr sction roxygen2
roxygen2/tests/testthat/test-tag-parser.R:42:    tag <- roxy_test_tag("foo params sction")
roxygen2/tests/testthat/_snaps/tag-parser.md:135:      tag <- roxy_test_tag("foo params sction")
roxygen2/tests/testthat/_snaps/tag-parser.md:139:      [test.R:1] @test attempts to inherit from unknown type "sction"
$ 
```

WON'T FIX

algorithm

```
$ grep -nr algoritm roxygen2
roxygen2/tests/testthat/_snaps/rd-inherit.md:4:    \item{algo}{The hashing algoritm to be used by \code{\link[digest]{digest}}. Defaults to
$ 
```

That typo is from the digest package documentation.

```
https://journal.r-project.org/archive/2012-1/RJournal_2012-1_Hornik~et~al.pdf

R uses the “Relator and
Role” codes and terms (http://www.loc.gov/marc/
relators/relaterm.html) from MARC (MAchine-
Readable Cataloging, Library of Congress, 2012), one
of the leading collections of standards used by librar-
ians to encode and share bibliographic information.
```

The MARC role `Redaktor` possibly is misspelled.

```
$ grep -nr redaktor roxygen2
[...]

PR #1501, opened 2023-07-07
fixed all typos listed in issue #1485

neshvig10:


neshvig10:
tests failed after I changed redaktor to redactor , hence swapped it back to the original spelling

PR #1520, opened 2023-10-19
Fix minor typo in reuse vignette

jonocarroll:
"`@describe` in" -> "`@describeIn`"


olivroy:
There is another open PR that aims to fix that (#1461).

PR #1528, opened 2023-11-04
qualify rlang ns in test

MichaelChirico:
Had the test suite fail for several reasons, one of which was 'caller_env not found'. It's in `roxygen2`'s namespace (via `import(rlang)`), but I think it's preferable to only rely on roxygen2's own objects in the tests without qualification/attachment.

### DIFF
--- a/tests/testthat/helper-test.R
+++ b/tests/testthat/helper-test.R
@@ -16,7 +16,7 @@ expect_parse_failure <- function(code)  {
   (expect_warning(expect_null(code)))
 }
 
-local_package_copy <- function(path, env = caller_env(), set_version = TRUE) {
+local_package_copy <- function(path, env = rlang::caller_env(), set_version = TRUE) {
   temp_path <- withr::local_tempdir(.local_envir = env)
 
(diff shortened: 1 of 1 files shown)


hadley:
Hmmmm, that feels weird to me. We expect that helpers have access to the package environment.

MichaelChirico:
I guess this stems from a "philosophical" issue I have with testthat operating inside the package namespace. I'll close at it seems a stylistic difference.

hadley:
FWIW I think it's pretty fundamental that testthat operate inside the package namespace since otherwise you can't test unexported functions.

MichaelChirico:
> FWIW I think it's pretty fundamental that testthat operate inside the package namespace since otherwise you can't test unexported functions.

right -- I don't think you should be able to do so :)

https://jesseduffield.com/Testing-Private-Methods/

hadley:
Wow!

MichaelChirico:
softening/clarification -- I think it should be _difficult_ to do so, especially requiring `:::` to distinguish public vs private API tests

PR #1534, opened 2023-11-16
Improve S3 method handling

### LINKED ISSUE #1175: Should roxygen warn when S3 methods are not tagged with `@export`?
It seems like it would help people with internal generics where it is not clear why you need to register. Cf r-lib/devtools#2293.

### LINKED ISSUE #1451: Improve documentation for S3 methods
Need to be more clear about how `export` works, and what to do if you want to export a method for a generic that's in a suggested package.

hadley:
Fixes #1175. Fixes #1451


hadley:
@jennybc while you're welcome to take a look at the warning implementation, I'd most appreciate your feedback on the changes to the vignette

vignettes/namespace.Rmd
```diff
@@ -19,53 +19,118 @@ This is a little frustrating at first, but soon becomes second-nature.
...
+  letters <- strsplit(x, "")
+  letters_rev <- lapply(letters, rev)
+  vapply(letters_rev, paste, collapse = "", FUN.VALUE = character(1))
+}
+```
+
+You have four options for documenting an S3 method:
+
+-   Don't document it; it's not required, and not needed for simple generics where the user won't care about the details.
+-   If the method is particularly complex or has many arguments that the generic does not, you can document it in its own file. In this case, just document it as if its a function.
```
jennybc, line 78:
```suggestion
-   If the method is particularly complex or has many arguments that the generic does not, you can document it in its own file. In this case, just document it as if it's a function.
```

vignettes/namespace.Rmd
```diff
@@ -19,53 +19,118 @@ This is a little frustrating at first, but soon becomes second-nature.
...
+-   You can use `@rdname` to document it with other methds for the generic. This is a good option if it's your generic, and you're providing a bunch of methods for different classes.
+-   You can use `@rdname` to document it with other methods for the class. This is typically the least appealing option because the different generics will have different arguments, leading to a cluttered and potentially confusing page.
+
+```{r}
+#' Take an object to bizarro world
+#' 
+#' @description
```
jennybc, line 85: We don't normally use `@description` explicitly. Is there a reason to do so in the current context?

vignettes/namespace.Rmd
```diff
@@ -19,53 +19,118 @@ This is a little frustrating at first, but soon becomes second-nature.
...
+bizarro.character <- function(x, ...) {
+  letters <- strsplit(x, "")
+  letters_rev <- lapply(letters, rev)
+  vapply(letters_rev, paste, collapse = "", FUN.VALUE = character(1))
+}
+```
+
+Typically, you will write methods for generics that are either defined in the current package or a package that is a hard dependency[^2] of your package.
+Sometimes, however, you will want to write a method for a suggested dependency.
+In this case, `@export` will not work because it assumes the generic is included or imported in your `NAMESPACE`. Instead, use `@exportS3Method`. This will use "delayed" method registration, which means the method will only be registered when the suggested package is loaded.
```
jennybc, line 109: So do we need to update this in R Packages 🤔 ? Scroll to the end of this section:

https://r-pkgs.org/dependencies-in-practice.html#imports-and-exports-related-to-s3

vignettes/namespace.Rmd
hadley: Because I think you want the bulleted listed of methods above the fold?  (i.e. we do use an explicit `@description` when there's a list)

vignettes/namespace.Rmd
hadley: Yeah, we should. We'll probably also want to sunset that `s3_register()` method in our own packages once we can depend on 3.6.  I'll file some issues so we don't forget.

PR #1535, opened 2023-11-16
Make `@docType package` work more like `_PACKAGE`

### LINKED ISSUE #1491: broken `@docType package` documentation
**roxygen2** 7.0.0 (2019-11-12) had the breaking change

> * Using `@docType package` no longer automatically adds a `-package` alias.
>   Instead document `_PACKAGE` to get all the defaults for package documentation

from <https://github.com/r-lib/roxygen2/commit/29db38fb823fcfe9cf2ad05f885bf20dfa1c964b>.

However, thousands of packages still use the old way ([that was still recommended a year ago](http://web.archive.org/web/20220607010527/https://r-pkgs.org/man.html#man-packages))
```
#' @docType package
#' @name pkgname
NULL
```
not being aware that they should

- either document the new ["special sentinel" `_PACKAGE`](https://r-pkgs.org/man.html#sec-man-package-doc)
- or manually add `@aliases <pkgname>-package`

because they no longer automatically get that alias that is required for a package overview page to be found via `package?pkgname`. As a consequence, CRAN now has thousands of packages where this help request no longer works.

Could **roxygen2** warn users when there is a `@docType package` block without corresponding `@aliases` (and possibly recommend the new `_PACKAGE` feature)?

hadley:
While recommending that you update to the newer syntax.

Fixes #1491


bastistician:
In principle, multiple help pages could be classified with `\docType{package}`, not only the `PKG-package` overview page. For example, I have seen packages that set this type also for `PKG-deprecated` or `PKG-defunct` overview pages, justifiably, even if the docType info is currently not used much by R itself ... So I am not sure if it makes sense to both deprecate and expand `@docType package` towards `"_PACKAGE"`. OTOH, also roxygen2 users can always create additional Rd files directly if needed.

I haven't looked at the internals, but maybe a less disruptive alternative could be for `@docType package` to only make sure that there is a PKG-package alias if also `@name PKG` (almost restoring the pre-7.0.0 behavior), and only in this case (where the desired alias was not set) throw a message about the new `"_PACKAGE"` feature (so people can opt out).

hadley:
While in principle, there might be multiple topics with `\docType{package}`, I don't think there's any practical advantage to it? As far as I know there is no special treatment for these topics.

PR #1538, opened 2023-11-17
Special case imports code to run first

### LINKED ISSUE #1144: devtools::document() errors before NAMESPACE can be updated
When running `devtools::document()` using roxygen2 to manage the namespace, the function errors before the NAMESPACE file can be updated/written.

I have run into this in two situations:

(1) adding a new import; e.g. trying to add the `S4Vectors::setValidity2` function using `#' @importFrom S4Vectors setvalidity2`:

    devtools::document()
    # Updating badInstall documentation
    # Loading badInstall
    # Error in setValidity2(Class = "TestClass", method = .TestClass.validity) 
    # (from methods-TestClass.R#12) : 
    #   could not find function "setValidity2"

(2) renaming a generic method:

    devtools::document()
    # Updating badInstall documentation
    # Loading badInstall
    # Error in add_classes_to_exports(ns = nsenv, package = package, exports = exports,  : 
    #   in ‘badInstall’ methods for export not found: myOldGeneric
    #   In addition: Warning message:
    #   In setup_ns_exports(path, export_all, export_imports) :
    #     Objects listed as exports, but not present in namespace: myOldGeneric

I am using RStudio: v1.3.959

    sessionInfo()
    # R version 4.0.0 (2020-04-24)
    # Platform: x86_64-apple-darwin17.0 (64-bit)
    # Running under: macOS Catalina 10.15.4
    # 
    # Matrix products: default
    # BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
    # LAPACK: /Library/Frameworks/R.framework/Versions/4.0/Resources/lib/libRlapack.dylib
    # 
    # locale:
    # [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
    # 
    # attached base packages:
    # [1] stats     graphics  grDevices utils     datasets  methods   base     
    # 
    # loaded via a namespace (and not attached):
    #  [1] Rcpp_1.0.4.6        compiler_4.0.0      prettyunits_1.1.1  
    #  [4] remotes_2.1.1       tools_4.0.0         testthat_2.3.2     
    #  [7] digest_0.6.25       pkgbuild_1.0.8      pkgload_1.0.2      
    # [10] memoise_1.1.0       rlang_0.4.6         cli_2.0.2          
    # [13] rstudioapi_0.11     parallel_4.0.0      xfun_0.14          
    # [16] withr_2.2.0         stringr_1.4.0       roxygen2_7.1.0     
    # [19] knitr_1.28          xml2_1.3.2          desc_1.2.0         
    # [22] S4Vectors_0.26.1    fs_1.4.1            devtools_2.3.0     
    # [25] stats4_4.0.0        rprojroot_1.3-2     glue_1.4.1         
    # [28] R6_2.4.1            processx_3.4.2      fansi_0.4.1        
    # [31] sessioninfo_1.1.1   callr_3.4.3         purrr_0.3.4        
    # [34] magrittr_1.5        backports_1.1.7     ps_1.3.3           
    # [37] ellipsis_0.3.1      BiocGenerics_0.34.0 usethis_1.6.1      
    # [40] assertthat_0.2.1    stringi_1.4.6       crayon_1.3.4

hadley:
Fixes #1144


R/namespace.R
```diff
@@ -71,27 +55,87 @@ roclet_clean.roclet_namespace <- function(x, base_path) {
...
+  "import",
+  "importFrom",
+  "importClassesFrom",
+  "importMethodsFrom",
+  "useDynLib"
+)
+
+update_namespace_imports <- function(base_path) {
+  NAMESPACE <- file.path(base_path, "NAMESPACE")
+  if (!made_by_roxygen(NAMESPACE) || !file.exists(NAMESPACE)) {
```
MichaelChirico, line 70: doesn't `made_by_roxygen()` run `file.exists()`? Either way it seems odd to run `file.exists()` second here

hadley: Isn't the logic opposite? We treat a missing `NAMESPACE` as made by roxygen2 so we can override it. Or are you just suggesting I flip the order of the conditional?

MichaelChirico: I'm a bit torn... from readability, it makes more sense to me like (1) check file exists (2) check if it's made by roxygen, but in implementation IINM that would mean `file.exists()` is always run twice, whereas this ordering means it may only be run once.

`file.exists()` is of course not very expensive on most file systems so this is not v. important, sorry for the fly-by comment :)

hadley: No problem; I had to think this through too when I wrote it, which is probably a sign it needs some refactoring, but it doesn't quite hit the threshold to be worth it.

PR #1556, opened 2023-12-08
Add a clickable link to file when already exists and not generated by roxygen2

olivroy:
An alternative would also be to use 

```r
"Skipping {.path {path}}"
```

Using the approach in the [cli manual]( https://cli.r-lib.org/reference/links.html?q=file#link-text-1)

image

I removed the snapshot test and test for FALSE manually.


hadley:
Why remove the snapshot test?

olivroy:
It had a stochastic output. I could use `transform` to restore it, but I didn't know about it then!

hadley:
Oooh because it's a temp file.

olivroy:
I restored the snapshot test and fixed a bracket typo! This is ready for your review

hadley:
Perfect, thank you!

PR #1557, opened 2023-12-11
Non-syntactic S4 classes don't need backticks

hadley:
Fixes documentation issue in odbc


hadley:
Judging by odbc, I don't think this is correct 😬 But I can't figure out any way to make it work 😬

PR #1563, opened 2023-12-23
More careful ordering for `@family` tag

krlmlr:
Needed for DBI: `dbSendQuery()` vs. `dbSendQueryArrow()` .

Test-first, the second commit afbabca036e000124711ae4cd538167113bcaa07 shows the effect. Happy to revisit the implementation and/or switch the other tests to snapshots.

I suspect that the new `order_c()` should be used almost everywhere where we now have `order()` .


tests/testthat/test-rd-family.R
```diff
@@ -123,3 +123,29 @@ test_that("custom family prefixes can be set", {
 
   expect_match(out$get_value("seealso"), "^Custom prefix:")
 })
+
+test_that("careful ordering", {
+  # Can't use foo1 because it's used as class in another test
```
hadley, line 128: Is it an S4 class? It might need a `on.exit(removeClass("foo1"))` call.

hadley (approved):
Looks fine to me. I can get in the next roxygen2 release if you add a news bullet.

krlmlr:
Thanks, good to go from my end.

tests/testthat/test-rd-family.R
krlmlr: No, unfortunately, #1564 does not change the behavior here.

krlmlr: Happy to adapt if #1566 or a better version gets in.

krlmlr:
Done, finally.

PR #1564, opened 2024-01-02
Remove S4 class after test

krlmlr:
https://github.com/r-lib/roxygen2/pull/1563/files#r1439781552

PR #1565, opened 2024-01-03
Remove S4 class after test

krlmlr:
A correct version of #1564.


krlmlr:
Doesn't work because the new `"env"` attribute breaks tests.

PR #1566, opened 2024-01-03
Remove S4 class after test

krlmlr:
Finally, a correct version of #1564 that passes all tests. Ugly, but the only reasonable way I found.


hadley (approved):
Oh yeah because that code is run by roxygen2, not directly in the test.

PR #1571, opened 2024-01-14
Add a .devcontainer

MichaelChirico:
This allows users to develop roxygen2 directly from the browser with GitHub Codespaces.

One nice side effect is encapsulating (in the Dockerfile) the minimal setup needed to get roxygen2 prepared for development locally.

Config/Needs/development is a way to strike a balance between installing all of Imports+Suggests (takes longer to build the container) and missing some pieces essential to development (testthat's a good example -- you can't really do most dev work without running the test suite --> need testthat).

In {lintr}, we also added {pkgload} here since that is used a lot in dev workflow. ({roxygen2} already Imports {pkgload}, of course):

https://github.com/r-lib/lintr/blob/bad1632f23b77f41cec13fb6f1c883302f64f506/DESCRIPTION#L55

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -14,6 +14,7 @@
 ^\.github$
 ^pkgdown$
 ^\.covrignore$
+^\.devcontainer$
 ^azure-pipelines\.yml$
 ^\.Rprofile$
 ^r-packages$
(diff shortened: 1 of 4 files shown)


hadley:
Happy to include with the caveat that we won't maintain it, so if any problems crop up with it I'll be cc'ing you 😄

MichaelChirico:
makes sense+SGTM! hopefully it's of use to some others :)

PR #1573, opened 2024-01-14
Fix idempotency of roxygenize() for packages using multi-line @rawNamespace

### LINKED ISSUE #1572: multi-line @rawNamespace + @import directives lead to mangled NAMESPACE on re-run
As noted here:

https://github.com/r-lib/lintr/pull/2510

I've constructed a more minimal example with package:

```
DESCRIPTION file:
Package: myPkg
Version: 0.0.1
Encoding: UTF-8

R/foo.R:
#' @import stats
#' @rawNamespace
#' if (TRUE) {
#'   import(methods)
#' } else {
#'   import(utils)
#' }
#' @import graphics
NULL
```

`roxygenize()` works fine, with contents:

```r
# Generated by roxygen2: do not edit by hand


if (TRUE) {
  import(methods)
} else {
  import(utils)
}
import(graphics)
import(stats)
```

Attempt to `roxygenize()` with existing NAMESPACE fails:

```r
roxygen2::roxygenize()
# Setting `RoxygenNote` to "7.3.0.9000"
# Error in parse(path, keep.source = TRUE) : 
#   /tmp/myPkg/NAMESPACE:15:1: unexpected '}'
# 14: }
# 15: }
#     ^
```

Contents of `NAMESPACE` now:

```
# Generated by roxygen2: do not edit by hand


if (TRUE) {
  import(methods)
} else {
  import(utils)
}
  import(methods)
  import(utils)
if (TRUE) {
import(graphics)
import(stats)
}
} else {
```

MichaelChirico:
Closes #1572. NB: This includes #1571, which I used to write the update. I can excise it but leaving it in for now.

Tests still needed. Filing for discussion first on if this is the right approach.

There are two parts to the fix:

 1. Collapse the parsed entries to one line in `namespace_exports()`. This matches the output of `namespace_imports()`.
 2. Except for `namespace_imports()` reads in a leading `\n`, so we use `trimws()` too.

----

That's all well and good for the example we have in mind for #1572, but I do wonder if the issue is broader. That the `@rawNamespace` entry trips up `namespace_exports()` has a smell to it -- the `@rawNamespace` code is only (nested) _import_ directives, why is it being pulled in by `namespace_exports()`? OTOH we probably don't want to traverse the AST for these entries, which can in principle break the dichotomy of import vs. export directives (nothing prevents us from putting an `export()` entry in one branch and an `import()` entry in the other, or a mix of the two types in either/both branches). So maybe this is just a case of `namespace_exports()` being misleadingly named?


MichaelChirico:
Have a working test but it feels clunky, you might suggest something snappier :)

hadley:
IIRC the idea is that we can't easily tell if `@rawNamespace` is an import or an export, we should treat it as if it is. If it's not actually an import directive, it probably won't cause a problem.

But yeah, `namespace_exports()` is really `namespace_not_imports()` so that `union(namespace_imports(), namespace_exports())` gives you the complete namespace. 

Your suggested fix makes sense to me.

hadley:
Can you please merge/rebase and add a news bullet?

hadley (approved):

MichaelChirico:
> Can you please merge/rebase and add a news bullet?

Done -- still not sure the right approach for testing (as noted, the current test is no good because it actually passes on current `main`, i.e., independently of the change here).

One option is to add another test package, I just wonder if you see a way to get the regression test without needing to add a whole fake package.

hadley:
What if we just test the assertion that `namespace_exports()` returns one line for each expression?

MichaelChirico:
Hmm I worry that tests an implementation detail rather than being a regression test for the specific bug. For example, such a test could in the future keep working while due to refactor multi-line `@rawNamespace` as in the motivating example breaks again.

Added a minimal test package instead

R/namespace.R
```diff
@@ -109,12 +109,14 @@ namespace_imports_blocks <- function(srcref) {
...
 
+# NB: this is designed as the conjugate of namespace_imports(), so also
+#   includes @rawNamespace entries which may/may not also include import directives.
 namespace_exports <- function(path) {
   parsed <- as.list(parse(path, keep.source = TRUE))
 
   is_import_directive <- function(x) is_call(x, import_directives)
   export_lines <- attr(parsed, "srcref")[!map_lgl(parsed, is_import_directive)]
-  unlist(lapply(export_lines, as.character))
+  unlist(lapply(export_lines, function(x) paste(as.character(x), collapse = "\n")))
```
hadley, line 119:
```suggestion
  # Each multiline directives are a single element so they're sorted correctly
  unlist(lapply(export_lines, function(x) paste(as.character(x), collapse = "\n")))
```

hadley:
@MichaelChirico do you think it's worth a patch release for this?

hadley:
Ok, I'll add to my to do list.

PR #1574, opened 2024-01-15
Restore old behavior of @importFrom in some edge cases

### LINKED ISSUE #1570: Broken support for `#' @importFrom` with backticks not documented
roxygen2 7.3.0 fails for several of my packages. The only vaguely related NEWS entry I found is

> - `@importFrom` throws a friendlier error if you try and import a non-existing functions (@MichaelChirico, #1409).

Is the new behavior intended, is it worth waiting for a fix?

## roxygen2 7.3.0

``` r
library(roxygen2)

roc_proc_text(namespace_roclet(), "
  #' @importFrom purrr `%>%`
  NULL
  ")
#> ✖ <text>:2: @importFrom Excluding unknown export in from purrr: `` `%>%` ``.
#> [1] "importFrom(purrr,)"

roc_proc_text(namespace_roclet(), "
  #' @importFrom purrr %>%
  NULL
  ")
#> [1] "importFrom(purrr,\"%>%\")"
```

<sup>Created on 2024-01-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## roxygen2 7.2.3

``` r
library(roxygen2)

roc_proc_text(namespace_roclet(), "
  #' @importFrom purrr `%>%`
  NULL
  ")
#> [1] "importFrom(purrr,`%>%`)"

roc_proc_text(namespace_roclet(), "
  #' @importFrom purrr %>%
  NULL
  ")
#> [1] "importFrom(purrr,\"%>%\")"
```

<sup>Created on 2024-01-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico:
Closes #1570

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 * `@family` lists are now ordered more carefully, "foo1" comes after "foo" (#1563, @krlmlr).
 
+* `@importFrom magrittr "%>%"` works again (#1570, @MichaelChirico). Relatedly, `@importFrom`
+  directives not matching any known functions (e.g. `@importFrom utils plot pdf`) will not
+  produce invalid NAMESPACE files.
+
 # roxygen2 7.3.0
 
 ## New features
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -268,9 +268,13 @@ roxy_tag_ns.roxy_tag_importFrom <- function(x, block, env) {
   pkg <- x$val[1L]
   if (requireNamespace(pkg, quietly = TRUE)) {
     importing <- x$val[-1L]
-    unknown_idx <- !importing %in% getNamespaceExports(pkg)
+    # be sure to match '%>%', `%>%`, "%>%" all to %>% given by getNamespaceExports, #1570
+    unknown_idx <- !normalize_quotes(importing) %in% getNamespaceExports(pkg)
     if (any(unknown_idx)) {
       warn_roxy_tag(x, "Excluding unknown {cli::qty(sum(unknown_idx))} export{?s} in from {.package {pkg}}: {.code {importing[unknown_idx]}}")
+      if (all(unknown_idx)) {
+        return(NULL)
+      }
       x$val <- c(pkg, importing[!unknown_idx])
     }
   }
--- a/R/utils.R
+++ b/R/utils.R
@@ -179,3 +179,4 @@ auto_quote <- function(x) {
 
 is_syntactic <- function(x) make.names(x) == x
 has_quotes <- function(x) str_detect(x, "^(`|'|\").*\\1$")
+normalize_quotes <- function(x) str_replace(x, "^(`|'|\")(.*)\\1$", "\\2")
--- a/tests/testthat/test-namespace.R
+++ b/tests/testthat/test-namespace.R
@@ -395,6 +395,15 @@ test_that("can extract non-imports from namespace preserving source", {
 })
 
 test_that("Invalid imports throw a helpful error", {
+  # No matched functions --> no output
+  block <- "
+    #' @importFrom utils InvalidUtilsFunction
+    NULL
+  "
+  expect_snapshot(out <- roc_proc_text(namespace_roclet(), block))
+  expect_equal(out, character())
+
+  # Matched functions --> only drop unmatched functions
   block <- "
     #' @importFrom utils head InvalidUtilsFunction
     NULL
@@ -418,6 +427,23 @@ test_that("Invalid imports throw a helpful error", {
   expect_snapshot(out <- roc_proc_text(namespace_roclet(), block))
   expect_equal(out, "importFrom(AnUnknownUnavailablePackage,Unchecked)")
 
+  # make sure to match several forms of non-syntactic names
+  expect_no_warning(expect_equal(
+    roc_proc_text(namespace_roclet(), "#' @importFrom stringr %>%\nNULL"),
+    'importFrom(stringr,"%>%")'
+  ))
+  expect_no_warning(expect_equal(
+    roc_proc_text(namespace_roclet(), "#' @importFrom stringr '%>%'\nNULL"),
+    "importFrom(stringr,'%>%')"
+  ))
+  expect_no_warning(expect_equal(
+    roc_proc_text(namespace_roclet(), "#' @importFrom stringr `%>%`\nNULL"),
+    "importFrom(stringr,`%>%`)"
+  ))
+  expect_no_warning(expect_equal(
+    roc_proc_text(namespace_roclet(), '#\' @importFrom stringr "%>%"\nNULL'),
+    'importFrom(stringr,"%>%")'
+  ))
 })
 
 


jdblischak:
This bug also affects `#' @importFrom data.table ":="`. Might be worth mentioning in the news bullet and adding a test case

R/utils.R
```diff
@@ -179,3 +179,4 @@ auto_quote <- function(x) {
 
 is_syntactic <- function(x) make.names(x) == x
 has_quotes <- function(x) str_detect(x, "^(`|'|\").*\\1$")
+normalize_quotes <- function(x) str_replace(x, "^(`|'|\")(.*)\\1$", "\\2")
```
hadley, line 182: IMO this is stripping, not normalising. (I'd called it normalising if (e.g.) you were switching `` ` `` and `'` to `"`")

tests/testthat/test-namespace.R
```diff
@@ -418,6 +427,23 @@ test_that("Invalid imports throw a helpful error", {
   expect_snapshot(out <- roc_proc_text(namespace_roclet(), block))
   expect_equal(out, "importFrom(AnUnknownUnavailablePackage,Unchecked)")
 
+  # make sure to match several forms of non-syntactic names
+  expect_no_warning(expect_equal(
+    roc_proc_text(namespace_roclet(), "#' @importFrom stringr %>%\nNULL"),
+    'importFrom(stringr,"%>%")'
+  ))
```
hadley, line 434: This is a testing philosophy question — should you test `namspace_roclet()` for each of the quoting options or is it better to test one type of quotes for the whole flow and then rely on testing  `normalize_quotes()`/`strip_quotes()` for the others? I'd tend to think the later, but I'd love to know what you think.

R/namespace.R
```diff
@@ -268,9 +268,13 @@ roxy_tag_ns.roxy_tag_importFrom <- function(x, block, env) {
...
+    unknown_idx <- !normalize_quotes(importing) %in% getNamespaceExports(pkg)
     if (any(unknown_idx)) {
-      warn_roxy_tag(x, "Excluding unknown {cli::qty(sum(unknown_idx))} export{?s} in from {.package {pkg}}: {.code {importing[unknown_idx]}}")
+      warn_roxy_tag(x, "Excluding unknown {cli::qty(sum(unknown_idx))} export{?s} from {.package {pkg}}: {.code {importing[unknown_idx]}}")
+      if (all(unknown_idx)) {
+        return(NULL)
+      }
       x$val <- c(pkg, importing[!unknown_idx])
     }
   }
```
hadley, line 280: One more thought: how do the different types of quoting affect the `NAMESPACE`? Are they normalised elsewhere, or is it ok to use any of the forms without quoting?

hadley:
@jdblischak do you see something systematically different in that use case to the magrittr example? Or do you think we should just make the news bullet a bit more clear that this applies to any non-syntactic function name?

jdblischak:
> Or do you think we should just make the news bullet a bit more clear that this applies to any non-syntactic function name?

Exactly. From this PR and the linked Issue, it gives the impression that the edge case is specific to the magrittr pipe.

Also, I just learned that `@importFrom dplyr %>%` is valid with both roxygen2 7.2.3 and 7.3.0. In the past I've always put non-syntactic function names in quotes. Either I was being overly cautious, or this used to be necessary in older versions. But if using quotes is an edge case, the news bullet could also inform users that the quotes are unnecessary

R/namespace.R
MichaelChirico: Do you mean from roxygen2 perspective or base R?

Answering for the latter: NAMESPACE needs to parse as R, so `importFrom(magrittr, %>%)` won't work. 

So for roxygen2, my understanding is `@importFrom magrittr %>%` will add quotes for convenience (that's the helper `auto_quotes`), but quoted forms are dropped in as-is.

tests/testthat/test-namespace.R
MichaelChirico: I lean to be as verbose/comprehensive as possible, even if it's very obvious from looking at the _current_ code that we're covered. My thinking for adding all of them is it guards against any eventual refactoring that forgets to cover all four options for quoting.

One thing we use to reduce to repetitiveness is condense things into a [{patrick}](https://cran.r-project.org/web/packages/patrick/index.html) test.

MichaelChirico:
@jdblischak thanks for the feedback, it's a great point we can use this bullet to remind readers of available convenience features. PTAL if the current bullet gets the point across.

NEWS.md
```diff
@@ -2,6 +2,12 @@
 
 * `@family` lists are now ordered more carefully, "foo1" comes after "foo" (#1563, @krlmlr).
 
+* `@importFrom magrittr "%>%"` works again, as does
+  ``@importFrom rlang `:=` `` (#1570, @MichaelChirico). The form
```
hadley, line 6: I think I'd start the bullet with something like `` `@importFrom` with quoted non-syntactic functions works again, e.g. ...`.

tests/testthat/test-namespace.R
hadley: Patrick looks interesting, although I immediately want to tweak the syntax from:

```R
with_parameters_test_that("Data is successfully converted:", {
    input <- convert(test_data)
    expect_type(input, type)
  },
  test_data = list(numeric_data, character_data),
  type = c("double", "character"),
  .test_name = type
)
```

to:

```R
test_these("{type} data is successfully converted:", 
  {
    input <- convert(test_data)
    expect_type(input, type)
  },
  list(
    double = list(test_data = numeric_data, type = "double"),
    character = list(test_data = character_data, type = "character")
  )
)
```

I do wonder if testthat should incorporate patrick-like functionality since it is a pretty common need.

hadley (approved):

tests/testthat/test-namespace.R
MichaelChirico: cc @michaelquinn32. I do see some old threads in {testthat}: #962, #563. Should we revive one of those, or start a new issue?

`test_these()` does sound pretty nice as a function name. Glad you found https://github.com/google/patrick/issues/18 already, I was going to link that.

NEWS.md
MichaelChirico: Thanks for the feedback, PTAL.

jdblischak (approved):
Thanks @MichaelChirico!

tests/testthat/test-namespace.R
hadley: Let's keep talking in patrick, and see where we get.

tests/testthat/test-namespace.R
```diff
@@ -394,32 +394,63 @@ test_that("can extract non-imports from namespace preserving source", {
   expect_equal(namespace_exports(path), lines[c(1:3, 5)])
 })
 
-test_that("Invalid imports throw a helpful error", {
+test_that("invalid imports generate correct declarations", {
```
hadley, line 397: @MichaelChirico I refactored into multiple tests since this one was getting a bit big.

tests/testthat/test-namespace.R
```diff
@@ -394,32 +394,63 @@ test_that("can extract non-imports from namespace preserving source", {
...
+    "#' @importFrom stringr %>%",
+    "#' @importFrom stringr `%>%`",
+    "#' @importFrom stringr '%>%'",
+    "#' @importFrom stringr \"%>%\"",
+    "NULL"
+  )
+
+  import <- expect_no_warning(roc_proc_text(namespace_roclet(), lines))
+  expect_equal(import, c(
+    "importFrom(stringr,\"%>%\")",
```
MichaelChirico, line 449: for my edification, any reason to prefer this to

```suggestion
    'importFrom(stringr,"%>%")',
```

I like this for keeping the 3 lines of code aligned vertically / no need to visually parse escapes.

tests/testthat/test-namespace.R
MichaelChirico: Looks good, thanks!

tests/testthat/test-namespace.R
hadley: I don't have particularly strong feelings here but I like that currently all the strings use `"`, and are consistent with the values in `lines`. I saw you used `'` for the input `"%>%"`, but I didn't like having to escape `#'`.

PR #1582, opened 2024-01-23
always use withAutoprint

### LINKED ISSUE #1581: Drop 'if getRversion() >= "3.4.0"' qualification under examplesIf tag?
https://github.com/r-lib/roxygen2/blob/db4dd9a4de2ce6817c17441d481cf5d03ef220e2/R/rd-examples.R#L20

[3.4.0 is now 7 years old](https://github.com/wch/r-source/tree/tags/R-3-4-0), and Tidyverse policy is to support R>=3.6.0.

Do we need to keep writing out this qualification on using `withAutoprint()` vs. `force()` indefinitely?

Obviously in the rendered Rd it doesn't make much difference, but we do occasionally need to look at the raw contents and it can be quite distracting in that case.

MichaelChirico:
Closes #1581

### DIFF
--- a/R/rd-examples.R
+++ b/R/rd-examples.R
@@ -15,9 +15,7 @@ roxy_tag_parse.roxy_tag_examplesIf <- function(x) {
   )
 
   dontshow <- paste0(
-    "\\dontshow{if (",
-    condition,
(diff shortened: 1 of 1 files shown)


hadley:
Looks good, thanks! Would you mind adding a news bullet?

hadley:
Also need to check if pkgdown needs an update for this too.

Issue #1585, opened 2024-01-30
S3 diagnostics suggests correction with typo


sven-stodtmann:
When missing `@exportS3Method`, I get the following, potentially very helpful message:
`S3 method method.cls needs @export or @exportS3method tag.`
However, when copy-pasting the suggestion, it leads to an error since "method" must be capitalized.
It should be
`S3 method method.cls needs @export or @exportS3Method tag.`
instead.

PR #1590, opened 2024-02-04
FIX @exportS3method -> @exportS3Method

HenrikBengtsson:


hadley:
Can you motivate this change for me? (Apologies if you already have but I've forgotten)

olivroy:
@hadley The tag name is `@exportS3Method`, but some interactive advice mentions the non-existing `@exportS3method` tag. This PR would fix #1585, #1591

hadley:
Ah got it, thanks!

MichaelChirico:
+1 just hit this to. I copy-pasted the `exportS3method` in the error and was all the more confused when the error became "unknown tag".

Maybe it makes sense to add `exportS3method` as an alias as well? capitalization doesn't seem all that important here...

PR #1594, opened 2024-02-09
Improve check for unexported S3 methods

### LINKED ISSUE #1593: Checking for missing S3 method exports hangs "forever" if you have large data objects in the package
In package yet unpublished I have a large data object (the `.rda` file has about 6 MB). With current roxygen versions, the namespace roclet hangs here

https://github.com/r-lib/roxygen2/blob/main/R/namespace.R#L402

This is apparently very inefficient if you have large objects in `s3objects`. If I remove the data object from the package, namespace generation proceeds swiftly as usual.

jranke:
Fixes #1593


hadley:
Thank you! Do you want to add a news bullet?  It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jranke:
Good point, I added a NEWS entry as suggested.

hadley:
Thank you!

PR #1598, opened 2024-03-06
FIX: Use exportS3Method instead of exportS3method

MLopez-Ibanez:


MLopez-Ibanez:
Actually, PR #1590 fixes this more thoroughly

PR #1604, opened 2024-04-08
Use more generic `x-r-run` over `ide:run`

DavisVaughan:
I was going to switch us away from `style_hyperlink()` entirely in favor of `format_inline("{.run ...}")`, but it produces some weird text when in the fallback case

- `style_hyperlink()` returns `text` unchanged in the fallback case
- `.run {text}` returns `text` styled as `{.code }` in the fallback case, i.e. it backticks it

So in the fallback case, if you combine `.run {text}` (backticks) with our usage of `{.path }` (single quotes) below, then you end up with ```` Writing '`a.Rd`' ````. It seems better to just continue to manually use `style_hyperlink()` so we have complete control over the fallback view. This case is covered by a test, which is how I learned about it in the first place.

https://github.com/r-lib/roxygen2/blob/db4dd9a4de2ce6817c17441d481cf5d03ef220e2/R/utils.R#L90-L93

### DIFF
--- a/R/rd.R
+++ b/R/rd.R
@@ -60,14 +60,14 @@ roclet_output.roclet_rd <- function(x, results, base_path, ..., is_first = FALSE
 
   names <- unname(map_chr(results, ~ .$get_name()[[1]]))
   if (length(names) > 0) {
-    hrefs <- paste0("ide:run:pkgload::dev_help('", names, "')")
+    commands <- paste0("pkgload::dev_help('", names, "')")
   } else {
-    hrefs <- character()
+    commands <- character()
   }
 
   # Always check for roxygen2 header before overwriting NAMESPACE (#436),
   # even when running for the first time
-  mapply(write_if_different, paths, contents, href = hrefs)
+  mapply(write_if_different, paths, contents, command = commands)
 
   if (!is_first) {
     # Automatically delete any files in man directory that were generated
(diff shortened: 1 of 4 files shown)


DavisVaughan:
The test failures look unrelated but do look like some kind of regressions that should probably be figured out first, in particular this error message looks worse all of a sudden

```
Last 13 lines of output:
  -   Quitting from lines 2-3 [unnamed-chunk-2] (<temp-path.Rmd>)
  +   Quitting from lines  at lines 2-3 [unnamed-chunk-2] (<temp-path.Rmd>)
      
  -   Quitting from lines 2-3 [unnamed-chunk-1] (<temp-path.Rmd>)
  +   Quitting from lines  at lines 2-3 [unnamed-chunk-1] (<temp-path.Rmd>)
      x <text>:3: @includeRmd failed to evaluate '<temp-path.Rmd>'.
      Caused by error:
      ! Error
  
  * Run `testthat::snapshot_accept('rd-include-rmd')` to accept the change.
  * Run `testthat::snapshot_review('rd-include-rmd')` to interactively review the change.
```

i.e. `Quitting from lines  at lines 2-3`

olivroy:
@DavisVaughan Caused by xfun. Solved in dev https://github.com/yihui/xfun/commit/1182e517306255873f3915ff98c75f0dcc2a1726

PR #1607, opened 2024-04-10
Add more inline markup

olivroy:
To make output more colorful

image

Snapshot failures caused by xfun (will be resolved with 0.44) (already solved in dev https://github.com/yihui/xfun/commit/1182e517306255873f3915ff98c75f0dcc2a1726


hadley:
Thank you!

PR #1608, opened 2024-04-10
Support Markdown in custom `@family` titles

salim-b:
Parses [custom `@family` titles](https://roxygen2.r-lib.org/articles/index-crossref.html) as Markdown.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # roxygen2 (development version)
 
+* Custom [`@family`
+  titles](https://roxygen2.r-lib.org/articles/index-crossref.html) now support
+  Markdown syntax (#1608, @salim-b).
+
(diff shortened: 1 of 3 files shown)


hadley:
Do the family name also end up in the concept tag?

salim-b:
> Do the family name also end up in the concept tag?

Sorry, I don't understand the question... what are you exactly referring to with "the concept tag"?

AFAICT, the `tag` arg given to `markdown()` is only used to create better error and warning msgs...

hadley:
I think family names also end up in `\concept{}`

salim-b:
> I think family names also end up in `\concept{}`

Yes, the family *name*, not the value (for which this PR adds Markdown support).

hadley:
Ah great.

PR #1611, opened 2024-04-19
Fix 3 typos in @describeIn section of the "Reusing documention" vignette

kellijohnson-NOAA:
This is a very minor pull request that only fixes three small typos that I found while reading the helpful vignette.


hadley:
Thank you!

PR #1615, opened 2024-04-27
avoid depend on the object subset method (`$`)

Yunuuuu:
It would be nice if R6-like documents can be used by other similar object.

I want to dispath methods written in rust and then write R document with roxygen2.

Since I will re-dispath `$.` and `[[.` methods of the namespace, which will cause roxygen2 not working. This commit just avoid `$.` and `[[.` methods when dealing with R6 object document by using `.subset2`, which should work for current all implementation.

Suppose I have an `Enum` object named `TimeUnit` from `rust`,  `Enum` object should be something similar with an environment bound with all active binding fields.

```r
new_namespace <- function(classname, public = list(),
                          active = list(), class = NULL) {
    # classname: used to define roxygen2 alias
    # class: add into final object class attribute
    if (missing(classname)) classname <- class
    if (!all_named(public) || !all_named(active)) {
        stop("All elements of public, and active must be named.")
    }
    if (!all_functions(active)) {
        stop("All items in active must be functions.")
    }
    namespace <- new.env(parent = emptyenv())

    # use the same name of `R6ClassGenerator` object.
    namespace$portable <- FALSE
    namespace$classname <- classname
    namespace$cloneable <- FALSE
    # namespace$class <- FALSE
    namespace$lock_class <- TRUE
    # https://github.com/r-lib/roxygen2/blob/main/R/object-r6.R#L139
    # namespace$inherit <- NULL # must be NULL for roxygen2 work
    # attr(namespace, "name") <- "rpolars_generator"

    # assign methods into namespace
    # namespace$private_fields <- NULL
    # namespace$private_methods <- NULL
    # namespace$public_fields <- NULL
    if (length(public)) {
        namespace$public_methods <- public
    }
    if (length(active)) {
        namespace$active <- active
    }

    # we assign a `R6ClassGenerator` class, in this ways,
    # we can utilize roxygen2 R6 style document
    structure(namespace, class = c(class, pkg_nm(), "R6ClassGenerator"))
}

# For example
# `call_timeunit_method` will call R `.Call` function.
# following will use R6 documents

#' TimeUnit object
tunits <- new_namespace(
    class = "tunits",
    active = list(
        #' @field Nanoseconds Unit of time "ns".
        Nanoseconds = function() {
            call_timeunit_method("Nanoseconds")
        },

        #' @field Microseconds Unit of time "us".
        Microseconds = function() {
            call_timeunit_method("Microseconds")
        },

        #' @field Milliseconds Unit of time "ms".
        Milliseconds = function() {
            call_timeunit_method("Milliseconds")
        }
    )
)

dispatch_method <- function(.__namespace__, .__class__) {
    # we use special name to prevent override other function (also known as private methods) defined this package
    force(.__namespace__)
    force(.__class__) # just for message
    function(self, .__name__) {
        # environment used to find `self`, `self` should be the data value
        # the parent environment of this environment is the package namespace
        # it's safe to just change the function environment in this package
        force(self)
        .__enclos_env__ <- environment()
        .__public_env__ <- .subset2(.__namespace__, "public_methods")
        .__active_env__ <- .subset2(.__namespace__, "active")
        if (!is.null(.__fn__ <- .subset2(.__public_env__, .__name__))) {
            environment(.__fn__) <- .__enclos_env__
            .__fn__
        } else if (
            !is.null(.__fn__ <- .subset2(.__active_env__, .__name__))) {
            environment(.__fn__) <- .__enclos_env__
            makeActiveBinding(".__active_fn__", .__fn__, .__enclos_env__)
            .__active_fn__ # nolint
        } else {
            stop(sprintf("No method `%s` found for %s", .__name__, .__class__))
        }
    }
}

#' @export
`$.tunits` <- dispatch_method(tunits, "`TimeUnit` namespace")

#' @export
`[[.tunits` <- `$.tunits`
```

### DIFF
--- a/R/object-r6.R
+++ b/R/object-r6.R
@@ -19,7 +19,7 @@ extract_r6_data <- function(x) {
 }
 
 drop_clone_maybe <- function(x, data) {
-  if (! "clone" %in% names(x$public_methods)) {
+  if (! "clone" %in% names(.subset2(x, "public_methods"))) {
     cline <- which(data$members$name == "clone" & data$members$type == "method")
     if (length(cline)) data$members <- data$members[-cline, ]
   }
@@ -39,13 +39,14 @@ default_r6_methods <- function() {
 }
 
 extract_r6_methods <- function(x) {
-  method_nms <- setdiff(names(x$public_methods), default_r6_methods())
+  public_methods <- .subset2(x, "public_methods")
+  method_nms <- setdiff(names(public_methods), default_r6_methods())
   method_loc <- map_int(
-    x$public_methods[method_nms],
+    public_methods[method_nms],
     function(m) {
       ref <- utils::getSrcref(m)
       if (is.null(ref)) {
-        name <- x$classname %||% "unknown"
+        name <- .subset2(x, "classname") %||% "unknown"
 
         cli::cli_abort(
           c(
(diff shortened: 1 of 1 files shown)


hadley:
I think this is too much of a big change to do without discussion and considering alternative approaches. Would you mind starting by filing an issue?

Issue #1619, opened 2024-05-07
The language is not respected when running inline code


olivroy:
My system language is French, but when I redocument packages, I'd like the language set in DESCRIPTION to be respected.

Given the following reprex

```r
library(roxygen2)
roc_proc_text(rd_roclet(), "
  #' Title
  #'
  #' ```{r}
  #' try(mean(x))
  #' ```
  #' @export
  foo <- function() {}"
)
```

```
$foo.Rd
% Generated by roxygen2: do not edit by hand
% Please edit documentation in ./<text>
\name{foo}
\alias{foo}
\title{Title}
\usage{
foo()
}
\description{
\if{html}{\out{<div class="sourceCode r">}}\preformatted{try(mean(x))
#> Error in eval(expr, envir, enclos) : objet 'x' introuvable
}\if{html}{\out{</div>}}
}
```

As you can see, the error is in French. `objet 'x' introuvable`

When I redocument rlang, I get some changes in the Rd file, so I have to be careful not to commit them. So I wonder if roxygen2 could set the locale internally so that this doesn't happen

hadley:
Unfortunately I feel like this is a high-risk manoeuvre for a relatively small gain. We know that setting language is tricky, and can have unintended consequences. Given that this problem affects few people and there are existing work arounds (i.e. you can call `local_language()` yourself), I don't think this is worth changing.

PR #1622, opened 2024-05-09
Set language to english in `local_reproducible_output()`

olivroy:
Addresses #1619.

Probably need to find a better way to respect language set in DESCRIPTION, any way to access this?

Similar to the approach in #1621

### DIFF
--- a/R/rd-eval.R
+++ b/R/rd-eval.R
@@ -23,5 +23,9 @@ local_reproducible_output <- function(.envir = parent.frame()) {
     .local_envir = .envir
   )
   withr::local_envvar(RSTUDIO = NA, .local_envir = .envir)
+  if (isTRUE(capabilities("NLS")) && Sys.getenv("LANG") !=
+      "C") {
+    withr::local_language("en", .local_envir = .envir)
+  }
   withr::local_collate("C", .local_envir = .envir)
 }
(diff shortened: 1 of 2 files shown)


IndrajeetPatil:
Something like this?

``` r
setwd("/Users/indrajeetpatil/Documents/GitHub/lintr")
desc::desc_get("Language")
#> Language 
#>  "en-US"

setwd("/Users/indrajeetpatil/Documents/GitHub/styler")
desc::desc_get("Language")
#> Language 
#>       NA
```

<sup>Created on 2024-05-10 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

R/rd-eval.R
```diff
@@ -23,5 +23,11 @@ local_reproducible_output <- function(.envir = parent.frame()) {
     .local_envir = .envir
   )
   withr::local_envvar(RSTUDIO = NA, .local_envir = .envir)
+
+  if (isTRUE(capabilities("NLS")) && Sys.getenv("LANG") != "C") {
+    lang <- unname(desc::desc_get("Language")) %|% "en"
+    withr::local_language(lang, .local_envir = .envir)
+  }
```
olivroy, line 30:
```suggestion
  lang <- unname(desc::desc_get("Language")) %|% "en"
  withr::local_language(lang, .local_envir = .envir)
```

Actually, it should just set language all the times no need for the condition like it is in testthat.

hadley:
See discussion in linked issue; unfortunately I think this has too high a risk for unintended consequences.

PR #1627, opened 2024-06-07
Document that aliases should be space separated (not comma).

### LINKED ISSUE #1626: FR: document that `@aliases` should be separated by space in autocomplete
Info is here: https://roxygen2.r-lib.org/articles/index-crossref.html?q=aliases#aliases

Just add it here too image

olivroy:
Fix #1626. I have seen in the wild cases of persons who used commas for example.

https://github.com/r-spatial/sf/pull/2285

See the defunct file https://github.com/r-spatial/sf/pull/2285/files#diff-089405beae4488e32546bbf6c68166873b65f29323ca65f621de30eb80154e53 as an example.


hadley:
Hmmmm, I'd rather encourage people to put one alias on each line.

PR #1633, opened 2024-07-09
Resolve markdown links to external packages

### LINKED ISSUE #1612: Always resolve to "anchored" Rd xrefs
In order to improve cross-package links in the new single-page-html documentation, CRAN will start showing NOTEs for Rd pages which link to a topic that exists in multiple CRAN packages, such as `toJSON` (in jsonlite, RJSONIO, rjson).

Seemingly it is not trivial at CMD check time to resolve a topic to a given based on the package imports/suggests. They suggest you use "anchored" Rd links, which simply means using the full `\link[pkg:topic]{topic}` syntax.

If possible, way to reduce some of the fallout would be to resolve a `[topic]` name to the full `\link[pkg:topic]{topic}` at roxygenize time, by inspecting at the package namespace, i.e. to automatically determine the `pkg:` value when unset:

https://github.com/r-lib/roxygen2/blob/2357de6fe50df352f286a4ec77f65e00e582fa21/R/markdown-link.R#L12-L23

gaborcsardi:
- [x] ~~Anchor links to base packages as well.~~
- [x] [Do **not** anchor base packages.](https://github.com/r-lib/roxygen2/issues/1612#issuecomment-2218740209)
- [x] Tests.
- [x] News.
- [x] Docs.
- [x] Cache dependency info.
- [x] Resolve re-exported functions.

Closes #1612.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,10 @@
 * Custom [`@family`
   titles](https://roxygen2.r-lib.org/articles/index-crossref.html) now support
   Markdown syntax (#1608, @salim-b).
+
+* Unqualified markdown links to topics in external packages are now
+  automatically resolved (#1612).
+
 # roxygen2 7.3.2
 
 * `@includeRmd` now additionally sets `options(cli.hyperlink = FALSE)` to make
@@ -16,7 +20,7 @@
   (#1563, @krlmlr).
 
 * `@importFrom` works again for quoted non-syntactic names, e.g.
-  `@importFrom magrittr "%>%"` or ``@importFrom rlang `:=` `` 
+  `@importFrom magrittr "%>%"` or ``@importFrom rlang `:=` ``
   (#1570, @MichaelChirico). The unquoted form `@importFrom magrittr %>%`
   continues to work. Relatedly, `@importFrom` directives matching no known
   functions (e.g. `@importFrom utils plot pdf`) produce valid NAMESPACE files
@@ -29,11 +33,11 @@
 
 ## New features
 
-* `@docType package` now works more like documenting `"_PACKAGE"`, 
+* `@docType package` now works more like documenting `"_PACKAGE"`,
   creating a `{packagename}-package` alias and clearly suggesting that
   you should switch to `"_PACKAGE"` instead (#1491).
 
-* `_PACKAGE` will no longer generate an alias for your package name if 
+* `_PACKAGE` will no longer generate an alias for your package name if
   a function of the same name exists (#1160).
 
 * The NAMESPACE roclet now reports if you have S3 methods that are missing
@@ -41,32 +45,32 @@
   really registers the method) even if the generic is not. This avoids rare,
   but hard to debug, problems (#1175). You can suppress the warning with
   `@exportS3Method NULL` (#1550).
-  
-* The `NAMESPACE` roclet once again regenerates imports _before_ loading 
-  package code and parsing roxygen blocks. This has been the goal for a long 
-  time (#372), but we accidentally broke it when adding support for code 
+
+* The `NAMESPACE` roclet once again regenerates imports _before_ loading
+  package code and parsing roxygen blocks. This has been the goal for a long
+  time (#372), but we accidentally broke it when adding support for code
   execution in markdown blocks. This resolves a family of problems where you
   somehow bork your `NAMESPACE` and can't easily get out of it because you
   can't re-document the package because your code doesn't reload.
 
 ## Minor improvements and bug fixes
 
-* If you document a function from another package it is automatically 
-  imported. Additionally, if you set `@rdname` or `@name` you can opt out 
-  of the default `reexports` topic generation and provide your own docs 
+* If you document a function from another package it is automatically
+  imported. Additionally, if you set `@rdname` or `@name` you can opt out
+  of the default `reexports` topic generation and provide your own docs
   (#1408).
 
 * Generate correct usage for S4 methods with non-syntactic class names.
 
 * The `ROXYGEN_PKG` env var provides the name of the package being documented
   (#1517).
-  
-* `@describeIn foo` now suggests that you might want `@rdname` instead 
+
+* `@describeIn foo` now suggests that you might want `@rdname` instead
   (#1493). It also gives a more informative warning if you use it with an
   unsupported type (#1490).
 
-* In `DESCRIPTION`, URLs containing escapes in `URL` and `BugReports` are 
-  now correctly handled (@HenningLorenzen-ext-bayer, #1415). Authors can now 
+* In `DESCRIPTION`, URLs containing escapes in `URL` and `BugReports` are
+  now correctly handled (@HenningLorenzen-ext-bayer, #1415). Authors can now
   have multiple email addresses (@jmbarbone, #1487).
 
 * `escape_examples()` is now exported (#1450).
@@ -74,7 +78,7 @@
 * `@exportS3Method` provides the needed metadata to generate correct usage
   for S3 methods, just like `@method` (#1202).
 
-* `is_s3_generic()` now ignores non-function objects when looking for a 
+* `is_s3_generic()` now ignores non-function objects when looking for a
   candidate function. I believe this is closer to how R operates.
 
 * `@import` and friends are now ignored if they try to import from the
@@ -87,7 +91,7 @@
 * `@include` now gives an informative warning if you use a path that doesn't
   exist (#1497).
 
-* `@inherit` can now also inherit from `@format` (#1293). 
+* `@inherit` can now also inherit from `@format` (#1293).
 
 # roxygen2 7.2.3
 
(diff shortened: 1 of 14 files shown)


R/markdown-link.R
```diff
@@ -11,18 +11,27 @@
 #' ```
 #' MARKDOWN           LINK TEXT  CODE RD
 #' --------           ---------  ---- --
-#' [fun()]            fun()       T   \\link[=fun]{fun()}
-#' [obj]              obj         F   \\link{obj}
+#' [fun()]            fun()       T   \\link[=fun]{fun()} or
+#'                                    \\link[pkg::file]{fun()}
```
krlmlr, line 15:
```suggestion
#'                                    \\link[pkg:file]{fun()}
```

Similarly elsewhere?

R/markdown-link.R
```diff
@@ -174,6 +185,47 @@ parse_link <- function(destination, contents, state) {
...
+  if (me == "") return(NA_character_)
+
+  # if it is in the current package, then no need for package name, right?
+  if (has_topic(topic, me)) return(NA_character_)
+
+  # try packages in depends, imports, suggests first, error on name clashes
+  pkgdir <- pkgdir %||% roxy_meta_get("current_package_dir")
+  deps <- desc::desc_get_deps(pkgdir)
+  deps <- deps[deps$package != "R", ]
+  deps <- deps[deps$type %in% c("Depeneds", "Imports", "Suggests"), ]
```
krlmlr, line 200:
```suggestion
  deps <- deps[deps$type %in% c("Depends", "Imports", "Suggests"), ]
```

krlmlr (commented):
Works for me in duckdb, thank you for the very quick fix!

R/markdown-link.R
```diff
@@ -11,18 +11,27 @@
 #' ```
 #' MARKDOWN           LINK TEXT  CODE RD
 #' --------           ---------  ---- --
-#' [fun()]            fun()       T   \\link[=fun]{fun()}
-#' [obj]              obj         F   \\link{obj}
+#' [fun()]            fun()       T   \\link[=fun]{fun()} or
+#'                                    \\link[pkg:file]{pkg::fun()}
+#' [obj]              obj         F   \\link{obj} or
+#'                                    \\link[pkg:obj]{pkg::obj}
```
krlmlr, line 17: Not sure about this, but consistent with `[pkg::obj]` below:

```suggestion
#'                                    \\link[pkg:file]{pkg::obj}
```

R/markdown-link.R
```diff
@@ -174,6 +185,59 @@ parse_link <- function(destination, contents, state) {
   }
 }
 
+resolve_link_package <- function(topic, me = NULL, pkgdir = NULL) {
+  me <- me %||% roxy_meta_get("current_package")
```
krlmlr, line 189: Can `me` still be `NULL` here? I saw this when calling this function manually.

krlmlr (commented):
Super-minor stuff.

Next level: automatically add these qualifications with `@inheritParams` -- I have this in DBI, and I see how niche this probably is.

gaborcsardi:
@krlmlr 
> Next level: automatically add these qualifications with @inheritParams -- I have this in DBI, and I see how niche this probably is.

You mean adjusting the link target packages in Rd from `@inheritParams`? Doesn't that work already?

NEWS.md
```diff
@@ -20,7 +23,7 @@
   (#1563, @krlmlr).
 
 * `@importFrom` works again for quoted non-syntactic names, e.g.
-  `@importFrom magrittr "%>%"` or ``@importFrom rlang `:=` `` 
+  `@importFrom magrittr "%>%"` or ``@importFrom rlang `:=` ``
```
hadley, line 26: Looks like you accidentally stripped trailing whitespace.

R/markdown-link.R
```diff
@@ -11,18 +11,27 @@
...
+#'                                    \\link[pkg:file]{text}
 #' [text][pkg::fun()] text        F   \\link[pkg:file]{text}
 #' [text][pkg::obj]   text        F   \\link[pkg:file]{text}
-#' [s4-class]         s4          F   \\linkS4class{s4}
+#' [s4-class]         s4          F   \\linkS4class{s4} or
+#'                                    \\link[pkg:file]{s4}
 #' [pkg::s4-class]    pkg::s4     F   \\link[pkg:file]{pkg::s4}
 #' ```
 #'
+#' For the ones that have two RD variants the first version is used for
```
hadley, line 31:
```suggestion
#' For the links with have two `.Rd` variants the first version is used for
```

R/markdown.R
```diff
@@ -68,6 +70,7 @@ markdown <- function(text, tag = NULL, sections = FALSE) {
 #' @keywords internal
 
 markdown_pass1 <- function(text) {
+  rm(list = ls(envir = mddata), envir = mddata)
```
hadley, line 73: Don't we want to cache deps across all calls? In which case, `roxy_meta_get()` might be a better home for this cache?

krlmlr:
It changes things in tibble, trying in CI/CD now: https://github.com/tidyverse/tibble/actions/runs/9871695446/job/27260156746

Thank you very much!

R/markdown-link.R
```diff
@@ -174,6 +185,100 @@ parse_link <- function(destination, contents, state) {
...
+  me <- me %||% roxy_meta_get("current_package")
+  # this is  from the roxygen2 tests, should not happen on a real package
+  if (is.null(me) || is.na(me) || me == "") return(NA_character_)
+
+  # if it is in the current package, then no need for package name, right?
+  if (has_topic(topic, me)) return(NA_character_)
+
+  # try packages in depends, imports, suggests first, error on name clashes
+  pkgdir <- pkgdir %||% roxy_meta_get("current_package_dir")
+  deps <- mddata[["deps"]] <- mddata[["deps"]] %||% desc::desc_get_deps(pkgdir)
```
hadley, line 198: I found `mddata[["deps"]] <- mddata[["deps"]]` a bit hard to parse mentally.

R/markdown-link.R
```diff
@@ -174,6 +185,100 @@ parse_link <- function(destination, contents, state) {
...
+  deps <- deps[deps$package != "R", ]
+  deps <- deps[deps$type %in% c("Depends", "Imports", "Suggests"), ]
+  pkgs <- deps$package
```
hadley, line 201: Maybe we only need to cache `pkgs`?

R/markdown-link.R
```diff
@@ -174,6 +185,100 @@ parse_link <- function(destination, contents, state) {
...
+
+  pkg_has_topic <- pkgs[map_lgl(pkgs, has_topic, topic = topic)]
+  pkg_has_topic <- map_chr(pkg_has_topic, function(p) {
+    p0 <- find_reexport_source(topic, p)
+    if (is.na(p0)) p else p0
+  })
+  pkg_has_topic <- unique(pkg_has_topic)
+  base <- base_packages()
+  if (length(pkg_has_topic) == 1) {
```
hadley, line 210: Maybe explicitly handle `length(pkg_has_topic) == 0`  just to be clear?

R/markdown-link.R
```diff
@@ -174,6 +185,100 @@ parse_link <- function(destination, contents, state) {
...
+
+  # try base packages as well, take the first hit,
+  # there should not be any name clashes, anyway
+  for (bp in base) {
+    if (has_topic(topic, bp)) return(NA_character_)
+  }
+
+  cli::cli_abort(c(
+    "Could not resolve link to topic {.val {topic}} in the dependent
+     and base packages.",
```
hadley, line 231:
```suggestion
    "Could not resolve link to topic {.val {topic}} in the dependencies or base packages.",
```

hadley (approved):

krlmlr:
Works for tibble, green builds again!

NEWS.md
gaborcsardi: Thanks Positron! :/

R/markdown.R
gaborcsardi: It is fairly fast to do it once, so it is safer not to cache it across all calls, in case the dependencies change.

R/markdown-link.R
gaborcsardi: Yeah. But it is so much simpler than an `if () { } else {}`. Anyway, I can change it to an `if`.

gaborcsardi:
So, I have been using this for a bit, and it does not work very well when you are adding new functions that are not documented yet, or refactoring a package and renaming functions. The errors don't allow the changes to propagate properly.

One solution would be to have a warning instead of an error. This is easy to implement. The caveat is that you might end up with incorrect links, unless you run `document()` twice.

Another solution is to delay the resolution of the link target packages until the (new) topic of the documented package are known. This might be difficult to implement with how roxygen2 works currently. We would essentially need to do the final link fixup in the generated Rd docs.

krlmlr:
To better understand this: the problem is that this PR works worse than the main branch if documentation is added to a function that didn't have documentation before? What errors are you seeing?

hadley:
I think switching to warnings are probably a good start. Can we assume that if we can't find the topic it's in the current package?

gaborcsardi:
Good idea. R CMD check will catch the real errors anyway.

gaborcsardi:
@krlmlr Yeah, when you are refactoring or developing, you might have a state where you are referring to a function that is not documented. E.g. if you simply rename a function that you are cross-linking from a manual page. With this PR the package of the renamed function cannot be resolved, which causes an error, and the manual pages are not updated. So you are stuck in this state.

gaborcsardi:
@hadley OK, I turned them into warnings. Can you please take another look?

R/markdown-link.R
```diff
@@ -174,6 +185,100 @@ parse_link <- function(destination, contents, state) {
...
+  if (length(pkg_has_topic) == 0) {
+    # fall through to check base packages as well
+  } else if (length(pkg_has_topic) == 1) {
+    if (pkg_has_topic %in% base) {
+      return(NA_character_)
+    } else {
+      return(pkg_has_topic)
+    }
+  } else {
+    cli::cli_warn(c(
```
hadley, line 215: Can you use `warn_roxy_topic()` here (and below)?

R/markdown-link.R
```diff
@@ -174,6 +185,100 @@ parse_link <- function(destination, contents, state) {
...
+  # try base packages as well, take the first hit,
+  # there should not be any name clashes, anyway
+  for (bp in base) {
+    if (has_topic(topic, bp)) return(NA_character_)
+  }
+
+  cli::cli_warn(c(
+    "Could not resolve link to topic {.val {topic}} in the dependencies or base packages.",
+    "i" = paste(
+      "If you haven'r documented {.val {topic}} yet, or just changed its name, this is normal.",
```
hadley, line 230: I had a vague memory that roxygen2 maintains some index of topics for the current package, but I can't find any sign of it in the code, so I'm confident that this is the best you can do

hadley (approved):

R/markdown-link.R
gaborcsardi: I don't have the topic here I think. I am not even sure that we have a topic at this point. Maybe I could use `warn_roxy_tag()` if we pass the tag through the machinery?

hadley: Oh I pattern matched on the wrong `topic`.  I think it would be worth passing all the necessary info down to so `warn_roxy_tag()` can give a file name/line number.

R/markdown-link.R
gaborcsardi: Yeah, we don't have the topics yet here. This was basically my second idea above, i.e. wait until we have the topics and do the substitution after that. Probably possible, but I think the warnings are also ok.

R/markdown-link.R
gaborcsardi: This is was pretty easy in the end.

jeroen:
Great, this worked for all cases I tried!

jeroen (approved):

PR #1641, opened 2024-08-02
fix: Support multiline `@examplesIf` again

krlmlr:
Fallout from https://github.com/r-lib/roxygen2/pull/1582.

Symptoms: https://github.com/r-prof/profile/actions/runs/10209333641/job/28247195516#step:11:73

```
Error: Error in map(seq_along(tokens), function(i) { : ℹ In index: 7.
Caused by error in `map()` at roxygen2/R/block.R:216:3:
ℹ In index: 5.
Caused by error in `if (str_trim(x$raw) == "") ...`:
! the condition has length > 1
```

CC @MichaelChirico.

### DIFF
--- a/R/rd-examples.R
+++ b/R/rd-examples.R
@@ -15,10 +15,12 @@ roxy_tag_parse.roxy_tag_examplesIf <- function(x) {
   )
 
   x$raw <- paste(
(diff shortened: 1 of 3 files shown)


MichaelChirico:
thank you!

hadley:
Thanks!

PR #1664, opened 2024-10-12
More explicit on documenting data sets

### LINKED ISSUE #1663: Missing "dataset-name" in vignette?
https://github.com/r-lib/roxygen2/blob/9652d15221109917d46768e836eaf55e33c21633/vignettes/rd-other.Rmd#L23

IINM, this is about using `#'` blocks over the character string `"dataset-name"` to document a data set, however that string doesn't appear in the vignette?

c.f. https://github.com/tidyverse/tidyr/blob/622afe810825ffb5de750c060778b150c68ba71a/R/data.R#L33-L36

MichaelChirico:
Closes #1663


hadley:
Thanks!

PR #1685, opened 2024-11-26
Use local_package_copy() for test writing to test path

MichaelChirico:
Our test environment does not have write permissions, causing this to fail.

Usually I ignore these issues, but in this case {roxygen2} already has a relevant helper, it's just not being used in this case (probably because the test is of idempotency --> the test package is not expected to change at all).

I am proposing the change here for consistency sake, but it's not strongly motivated besides our use case -- feel free to close :)


hadley:
Thanks!

PR #1691, opened 2024-12-29
fix: allow returns tag to be used in r6 methods

m-muecke:
closes: https://github.com/r-lib/roxygen2/issues/1652, https://github.com/r-lib/roxygen2/issues/1148


hadley:
This doesn't look like the correct fix — it documents a return value for the entire object, not the specific method.

PR #1692, opened 2024-12-29
refactor: remove namespace prefix for purrr function that are imported

m-muecke:


hadley:
Thanks for taking a look at this, but I'd prefer to go in the opposite direction and reduce our dependence on purrr.

Issue #1589, opened 2024-02-02
roxygenize error: "Caused by error in `basename()`: a character vector argument expected"


andrew-schulman:
I've been successfully building my package documentation until now, up through roxygen2 7.2.3. But with version 7.3.0 and later, it fails:
```r
> roxygen2::roxygenize(roclets = c('rd','collate','namespace','vignette'))
ℹ Loading DMRI
Error in `map2()`:
ℹ In index: 1.
ℹ With name: log.debug.
Caused by error in `basename()`:
! a character vector argument expected
Run `rlang::last_trace()` to see where the error occurred.
> rlang::last_trace()
<error/purrr_error_indexed>
Error in `map2()`:
ℹ In index: 1.
ℹ With name: log.debug.
Caused by error in `basename()`:
! a character vector argument expected
---
Backtrace:
     ▆
  1. └─roxygen2::roxygenize(...)
  2.   └─base::lapply(...)
  3.     ├─roxygen2 (local) FUN(X[[i]], ...)
  4.     └─roxygen2:::roclet_process.roclet_namespace(X[[i]], ...)
  5.       └─roxygen2:::warn_missing_s3_exports(blocks, env)
  6.         └─purrr::map2(...)
  7.           └─purrr:::map2_("list", .x, .y, .f, ..., .progress = .progress)
  8.             ├─purrr:::with_indexed_errors(...)
  9.             │ └─base::withCallingHandlers(...)
 10.             ├─purrr:::call_with_cleanup(...)
 11.             └─roxygen2 (local) .f(.x[[i]], .y[[i]], ...)
 12.               └─roxygen2:::warn_roxy_function(fun, "S3 method {.arg {name}} needs @export or @exportS3method tag")
 13.                 └─roxygen2:::warn_roxy(file, line, message, parent = parent, envir = envir)
 14.                   ├─cli::style_hyperlink(...)
 15.                   │ └─base::paste0(...)
 16.                   ├─base::paste0(basename(file), ":", line)
 17.                   └─base::basename(file)
Run rlang::last_trace(drop = FALSE) to see 4 hidden frames.
```
Apparently one of my S3 methods is missing an `@export` tag, but since the warning message fails, I can't tell which one. Running again with `debug` helps to show what the trouble is:
```r
> debug(roxygen2:::warn_roxy_function) 
> roxygen2::roxygenize(roclets = c('rd','collate','namespace','vignette'))
Setting `RoxygenNote` to "7.3.1"
[...]

PR #1711, opened 2025-04-20
Fix #1710 Undocumented S3 Method Check Fails

### LINKED ISSUE #1710: Undocumented S3 Method Check Fails when srcrefs are Missing
Running `roxygenize` on a package with undocumented S3 methods and without srcrefs fails as follows:

```
Error in map2(undocumented, names(undocumented), function(fun, name) { : 
  ℹ In index: 1.
ℹ With name: c.blah.
Caused by error in `basename()`:
! a character vector argument expected
```

The error relates to the addition of checks for undocumented S3 methods from #1534 which implicitly assumes srcrefs are [always present](https://github.com/r-lib/roxygen2/blob/9652d15221109917d46768e836eaf55e33c21633/R/utils-warn.R#L17).  Downstream code then fails because it gets NULL instead of a string.

Recreating the error is a bit tricky b/c we need a package without srcrefs, but here is a reasonably small example that reproduces the error:

```
pkg.name <- 'pkguqhr'
DESCRIPTION <- sprintf('
Package: %s
Version: 0.0
Encoding: UTF-8
Roxygen: list(load="installed")', pkg.name)
NAMESPACE <- ''
RCODE <- 'c.blah <- function(...) NULL'
if(!pkg.name %in% rownames(installed.packages())) {
  pkg.dir <- tempfile(pkg.name)
  dir.create(file.path(pkg.dir, "R"), recursive=TRUE)
  writeLines(DESCRIPTION, file.path(pkg.dir, 'DESCRIPTION'))
  writeLines(NAMESPACE, file.path(pkg.dir, 'NAMESPACE'))
  writeLines(RCODE, file.path(pkg.dir, 'R', 'pkg.R'))

  install.packages(pkg.dir, type='src', repos=NULL, quiet=TRUE)
  try(roxygen2::roxygenise(pkg.dir))
  remove.packages(pkg.name)
} else {
  stop(
    "Please either remove ", pkg.name, 
    " or find another name for this test package"
  )
}
```

Notice the `Roxygen: list(load="installed")`.  Assuming the above is in `test.R` then:

```
R --vanilla -q -f test.R
```

Reproduces the error.

Relates to #1589.  #1711 provides a simple fix, but probably better would be to have an accessor function for srcrefs that is guaranteed to return in standard format, and then use that function everywhere, but that's a bigger change.

smilberg:
Fix #1710 and #1589.

### DIFF
--- a/R/utils-warn.R
+++ b/R/utils-warn.R
@@ -15,8 +15,8 @@ warn_roxy_block <- function(block, message, parent = NULL, envir = parent.frame(
 
 warn_roxy_function <- function(fun, message, parent = NULL, envir = parent.frame()) {
   srcref <- attr(fun, "srcref")
-  file <- attr(srcref, "srcfile")$filename
-  line <- as.vector(srcref)[[1]]
+  file <- attr(srcref, "srcfile")$filename %||% '<unknown-file>'
+  line <- as.vector(srcref)[[1]] %||% 0
 
   warn_roxy(file, line, message, parent = parent, envir = envir)
 }


smilberg:
FWIW my local R CMD check failed with the same error before and after this patch.

hadley:
Thanks for working on this! I merged upstream and then added a unit test and news bullet.

PR #1712, opened 2025-04-22
feat: Support markup in link text for R >= 4.5.0

salim-b:
From R 4.5.0 [NEWS](https://cran.r-project.org/doc/manuals/r-release/NEWS.html):

> - The Rd `⁠\link`⁠ macro now allows markup in the link text when the topic is given by the optional argument, e.g., `⁠\link[=gamma]{\eqn{\Gamma(x)}}⁠`.

This PR adapts roxygen2 internally to allow generating such Rd markup, see 6778edad64574f7c24c81394c3d5ab037f7086b3.

Additionally, it updates unrelated test snapshots (bd6ba1afae406d2f10834ca59f4c7cae2ef85a95) and includes some cosmetic changes (4fc0a3abc34b5049bc29df6611a187b2cd2de821).

The failing test (`resolve_link_package`) is unrelated to this PR and should be fixed by #1711.

### DIFF
--- a/R/markdown-link.R
+++ b/R/markdown-link.R
@@ -98,7 +98,8 @@ parse_link <- function(destination, contents, state) {
   if (!grepl("^R:", destination)) return(NULL)
   destination <- sub("^R:", "", URLdecode(destination))
 
-  ## if contents is a `code tag`, then we need to move this outside
+  ## If contents is a single \code tag, we move it outside
+  ## (still necessary in R >= 4.5 to support the short "[`obj_name`]" syntax)
   is_code <- FALSE
   if (length(contents) == 1 && xml_name(contents) == "code") {
     is_code <- TRUE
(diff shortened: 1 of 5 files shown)


hadley:
The unrelated diffs make it rather hard to see what you're doing here, so I'll take a swing at it myself.

salim-b:
> The unrelated diffs make it rather hard to see what you're doing here, so I'll take a swing at it myself.

https://github.com/r-lib/roxygen2/pull/1712/changes/6778edad64574f7c24c81394c3d5ab037f7086b3 is the core change. I've left a fallback for R < 4.5.0. Your #1757 is cleaner, but breaks backwards-compatibility. I'm fine with either.

hadley:
@salim-b per the PR comment, I don't think it does break backward compatibility in a meaningful way.

salim-b:
> @salim-b per the PR comment, I don't think it does break backward compatibility in a meaningful way.

Missed that, you're absolutely right.

Issue #1728, opened 2025-09-19
`forcats` `%in%` does not link


math-mcshane:
Hello from tidy dev day 2025, 

After running 

```r
usethis::create_from_github("tidyverse/forcats")
devtools::document()
```

I got the following message: 

```r
ℹ Updating forcats documentation
ℹ Setting RoxygenNote to "7.3.3"
ℹ Loading forcats
✖ match.R:3: @description Could not resolve link to topic "\\%in\\%" in the dependencies or base
  packages.
ℹ If you haven't documented "\\%in\\%" yet, or just changed its name, this is normal. Once "\\%in\\%"
  is documented, this warning goes away.
ℹ Make sure that the name of the topic is spelled correctly.
ℹ Always list the linked package as a dependency.
ℹ Alternatively, you can fully qualify the link with a package name.
Writing forcats-package.Rd
```

Examining `match.R`, @hadley noted that it appeared that roxygen2 did not link `%in%` and to leave this issue here.

```
#' Test for presence of levels in a factor
#'
#' Do any of `lvls` occur in `f`? Compared to [%in%], this function validates
#' `lvls` to ensure that they're actually present in `f`. In other words,
#' `x %in% "not present"` will return `FALSE`, but `fct_match(x, "not present")`
#' will throw an error.
```

Or [here](https://github.com/tidyverse/forcats/blob/b9049f7729c55b368525cbafd285b81ba89535b4/R/match.R#L1-L7).

mcol:
This started to occur with b94b42ff (#1633).

jmbarbone:
My workaround is just linking to `match()` instead: 

```r
#' [`%in%`][match()]
```

PR #1740, opened 2026-01-05
Hint about lacking support for links to operators

MichaelChirico:
Updating from {roxygen2} 7.3.2 to 7.3.3 generated a lot of logs like:

```
✖ integer64.R:171: @returns Could not resolve link to topic "\\%/\\%" in the dependencies or base packages.
ℹ If you haven't documented "\\%/\\%" yet, or just changed its name, this is normal. Once "\\%/\\%" is documented, this warning goes away.
ℹ Make sure that the name of the topic is spelled correctly.
ℹ Always list the linked package as a dependency.
ℹ Alternatively, you can fully qualify the link with a package name.
```

The current messaging is misleading / extra hard to parse due to the double-escaped `\`. It took some sleuthing to find that this is not supported: https://github.com/r-lib/roxygen2/issues/550. (I had checked the [Indexing and cross-references](https://roxygen2.r-lib.org/articles/index-crossref.html) vignette, not [(R)Markdown support]((R)Markdown support)).

Looks like this entails new shapshots -- can generate if this roughly looks good.

### DIFF
--- a/R/markdown-link.R
+++ b/R/markdown-link.R
@@ -232,7 +232,8 @@ resolve_link_package <- function(topic, me = NULL, pkgdir = NULL, state = NULL) 
       "Once {.val {topic}} is documented, this warning goes away."),
     "i" = "Make sure that the name of the topic is spelled correctly.",
     "i" = "Always list the linked package as a dependency.",
-    "i" = "Alternatively, you can fully qualify the link with a package name."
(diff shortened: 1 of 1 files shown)


hadley:
This tripped me up too (#1728) so I'd prefer to try and fix it properly.

PR #1747, opened 2026-02-03
Fixes #1746

### LINKED ISSUE #1746: Including arbitrary information into a the comment argument in a person object in authors@R in the DESCRIPTION fails
According to the documentation of `person()` we are allowed to include arbitrary information in the comment argument, which is specified as a character vector.

Printing such a person object also includes the names of the comments:

```
> person("Johannes", "Ranke", role = c("ctb"), comment = c(acronym = "rajo", "Contributed XY"))
[1] "Johannes Ranke [ctb] (acronym: rajo, Contributed XY)"
```

Currently, roxygen fails if the comment argument has length > 1 after special treatment of elements named ORCID and ROR. I think it should accept also other, arbitrary documentation, following the documentation of `person()`.

jranke:
The previous behaviour, putting ORCID and ROR information into an own set of parentheses, was not changed, e.g. we get

H W \\email{h@w.com} (\\href{https://orcid.org/0000-0003-4757-117X}{ORCID}) (extra)

It could be argued that there should only be one set of parentheses for comments, but I did not want to mess with this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # roxygen2 (development version)
 
+* For person defintions `DESCRIPTION`, the comment argument can now contain an arbitrary number of elements, and the names of these elements are included in the resulting string representation (@jranke, #1746)
+
 # roxygen2 7.3.3
 
 * Package documentation now converts ROR IDs into a useful link (#1698, @maelle).


hadley:
Oops, ended up fixing myself because I didn't spot your PR. For future reference, you need to include `Fixes #1746` in the description, not the title, in order to auto-link.

jranke:
No problem, your fix was nice enough😄. Thanks!

PR #1748, opened 2026-02-12
@inherit was mistakenly written as @inherits

JesseAlderliesten:
In the vignette about reusing documentation, the tag `@inherit` was mistakenly written as `@inherits`.
Using `@inherits` leads to the error `@inherits is not a known tag` when using `roxygen2::roxygenise()`.


hadley:
Thanks for the fix!

PR #1758, opened 2026-03-04
Improve link resolution

### LINKED ISSUE #1662: When resolving rmarkdown links to other packages, please do not anchor the link text.
Since https://github.com/r-lib/roxygen2/pull/1633 was merged, the behaviour has been to expand `[fun()]` to `\code{\link[pkg:fun]{pkg::fun()}}`.

This is often unsightly and verbose and is not ideal when the link text refers to a class rather than its main constructor. For example, ```[`mcmc.list`]``` becomes `\code{\link[coda:mcmc.list]{coda::mcmc.list}}`.

I think it would be better if the link text were only anchored if the link is anchored. E.g.,

`[fun()]` -> `\code{\link[pkg:fun]{fun()}}`
`[pkg::fun()]` -> `\code{\link[pkg:fun]{pkg::fun()}}`

### LINKED ISSUE #1724: Link resolution causes slowdown in 7.3.3
I've upgraded roxygen2 to 7.3.3 and running `roxygen2::roxygenise()` on the Luminescence package (https://github.com/R-Lum/Luminescence) went from 20 seconds to several minutes. At times (strangely enough, not always) I get tons of messages like this:
```
✖ Luminescence-generics.R:4: @description Could not resolve link to topic
  "RLum.Data-class" in the dependencies or base packages.
ℹ If you haven't documented "RLum.Data-class" yet, or just changed its name,
  this is normal. Once "RLum.Data-class" is documented, this warning goes
  away.
ℹ Make sure that the name of the topic is spelled correctly.
ℹ Always list the linked package as a dependency.
ℹ Alternatively, you can fully qualify the link with a package name.
```
I think I may be getting one such message for each link to any class or function defined in the package.

This may be similar to what is reported in #1720, but the analysis there doesn't seem to help in my case.

I've run a bisection, and it points to #1633. Indeed, commenting out the call to `resolve_link_package()` linked below removes the slowdown:
https://github.com/r-lib/roxygen2/blob/145e7b4d9aad3d063fa21b3da02139dede76a4e1/R/markdown-link.R#L150

### LINKED ISSUE #1725: [anyNA()] now resolving to backports::anyNA (etc)
Our CI to check man/ comments are in sync with the source code just started failing:

https://github.com/r-lib/lintr/actions/runs/17746754124/job/50433547555?pr=2919

```diff
-\code{\link[=anyNA]{anyNA()}} exists as a replacement for \code{any(is.na(x))} which is more efficient
+\code{\link[backports:anyNA]{backports::anyNA()}} exists as a replacement for \code{any(is.na(x))} which is more efficient
```

It seems something changed in {roxygen2} to cause this -- our most recent run on `main` 3 weeks ago didn't have this issue:

https://github.com/r-lib/lintr/actions/runs/17204746502/job/48802781565

The passing job, per the logs, was 7.3.2, while the failing one is 7.3.3.

We have {backports} in `Imports:`, but we don't import anything (e.g. any methods); currently, it's only used to import one object in .onLoad:

https://github.com/r-lib/lintr/blob/33efecf7f4aee58fbc9d4f2e1f3fda5ae908ed42/R/zzz.R#L311

So it seems to me it shouldn't resolve to the `?backports::anyNA` page.

### LINKED ISSUE #1728: `forcats` `%in%` does not link

hadley:
Lots of refactoring of code and tests

* Cache link resolution to avoid performance regression. Fixes #1724.  
* Include base packages in ambiguity detection. Fixes #1725. 
* Correctly resolve infix functions. Fixes #1728.
* Don't affect link text, only target. Fixes #1662.

@mcol for me this takes the documentation time of Luminescence from ~20s down to ~4.5s.

### DIFF
--- a/R/markdown-link-resolve.R
+++ b/R/markdown-link-resolve.R
@@ -1,18 +1,17 @@
 find_package <- function(topic, tag = NULL) {
-  pkg <- roxy_meta_get("current_package")
-  if (is.null(pkg)) {
+  cur_pkg <- roxy_meta_get("current_package")
+  cur_pkg_dir <- roxy_meta_get("current_package_dir")
+  if (is.null(cur_pkg)) {
     # Don't try and link in basic tests
     return(NA_character_)
   }
-  pkg_dir <- roxy_meta_get("current_package_dir")
 
-  pkg <- find_package_cached(topic, pkg = pkg, pkg_dir = pkg_dir)
-
+  pkg <- find_package_cached(topic, pkg = cur_pkg, pkg_dir = cur_pkg_dir)
   if (length(pkg) == 0) {
     warn_roxy_tag(
       tag,
       c(
-        "Could not resolve link to topic {.val {topic}} in the dependencies or base packages",
+        "Could not resolve link to topic {.val {topic}} in the dependencies or base packages.",
         "i" = paste(
           "If you haven't documented {.val {topic}} yet, or just changed its name, this is normal.",
           "Once {.val {topic}} is documented, this warning goes away."
@@ -29,7 +28,7 @@ find_package <- function(topic, tag = NULL) {
     warn_roxy_tag(
       tag,
       c(
-        "Topic {.val {topic}} is available in multiple packages: {.pkg {pkg}}",
+        "Topic {.val {topic}} is available in multiple packages: {.pkg {pkg}}.",
         i = "Qualify topic explicitly with a package name when linking to it."
       )
     )
@@ -38,8 +37,8 @@ find_package <- function(topic, tag = NULL) {
 }
 
 find_package_cache <- new_environment()
+# run in roxygenize() because the documented functions might change between runs
 find_package_cache_reset <- function() {
-  # run in roxygenize()
   env_unbind(find_package_cache, env_names(find_package_cache))
 
   # also reset the cache used by dev_help() (used by has_topic())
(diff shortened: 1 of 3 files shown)


mcol:
Fantastic! For me on Luminescence, I go from 28s to 7s. :)

MichaelChirico:
FTR my own speed-up ({bit64}) is more modest: 15s 📉 10s

PR #1780, opened 2026-03-06
S3 methods now match longest generic name first

### LINKED ISSUE #1587: `@export` all.equal S3 method error
Having an entry like this:
```r
#' @export
all.equal.keras.src.backend.common.variables.KerasVariable <-
  function(target, current, ...) { }
```

Produces this in the NAMESPACE
```
S3method(all,equal.keras.src.backend.common.variables.KerasVariable)
```
It should match `all.equal`, not `all`.

Seems like a return of https://github.com/r-lib/roxygen2/issues/147

hadley:
Previously we matched the shorted generic name first, which lead to `all.equal.abc` being interpreted as the `equal.abc` method for the `all` generic. This will now return the wrong result if you really did make an `equal.abc` class, but I think that should be much much less common.

Fixes #1587

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,13 +1,17 @@
 # roxygen2 (development version)
 
+* `@examples` no longer warns about unmatched braces inside of raw strings (#1492).
+* `@examples` no longer warns about unmatched braces when braces appear inside strings within R comments, e.g. `# '{greeting}'` (#1492).
 * `@inheritParams` now correctly updates `\linkS4class{}` links when inheriting parameter documentation from other packages, converting them to absolute links (#1634).
 * Package documentation now lists a person with both `"aut"` and `"cre"` roles in both the Maintainer and Authors sections (#1588).
 * Markdown links now do a better job of resolving package names: the process is cached for better performance (#1724); it works with infix operators (e.g. `[%in%]`) (#1728); no longer changes the link text (#1662); and includes base packages when reporting ambiguous functions (#1725).
 * Package documentation now uses `logo.svg` if available, falling back to `logo.png` (#1640).
 * `@family` no longer adds a trailing space after the colon in the default family prefix (#1628). Custom `rd_family_title` values now automatically get a colon appended if they don't already end with one (#1656).
+* `@inheritDotParams` now generates an informative warning when the source function can't be found, instead of a cryptic error (#1602).
 * `@inheritDotParams` now warns and produces no output when there are no parameters to inherit, instead of generating an empty `\describe` block that caused CRAN HTML validation warnings (#1671).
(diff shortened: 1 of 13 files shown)


MichaelChirico:
FWIW, I regex searched all CRAN for potential ambiguities against {base} S3 registrations (i.e., those in the default `.__S3MethodsTable__.`)

I _only_ found cases where the new rule is an improvement (and almost all of those are the `all,equal.<class>` mistake from the bug). For example:

 - [cases where the true generic uses `.`](https://github.com/search?q=org%3Acran+path%3ANAMESPACE+%2FS3method%5C%28.*%28%5B%5Et2%5D%7C%5B%5Es%5Dt%7C%5B%5Ea%5Dst%7C%5B%5Ec%5Dast%7C%5B%5Ee%5D2%7C%5B%5Ep%5De2%29%2C%28character%7Cenvironment%7Cna%7Cnan%7Clogical%7Cfinite%7Cinfinite%7Cnumeric%7Carray%7Ctable%7Cvector%7Cexpression%7CDa+te%7CPOSIX%7Cdouble%7Cnull%7Cfunction%7Csingle%7Cmatrix%7Clist%29%5B.%5D%2F+-repo%3Acran%2Fsantoku+-repo%3Acran%2Fmemisc&type=code)
 - [cases where the true class uses `.`](https://github.com/search?q=org%3Acran+path%3ANAMESPACE+%2FS3method%5C%28.*%5B.%5D%28decomposed%7Cpairwise%7Cpower%7Cprofile%7Csmooth%7Csummary%7Csimple%29%2C%2F+-repo%3Acran%2Fsantoku+-repo%3Acran%2Fmemisc&type=code)

hadley:
@MichaelChirico thanks for the analysis!

hadley:
Oh but it totally breaks our hacky vctr's double dispatch mechanism :(

MichaelChirico:
Hm I looked at {vctrs} -- I thought all style guide-compliant code would be immune because it should be `snake_case`, therefore no generic uses `.` --> no ambiguity?

hadley:
Hmm, I'll have to look into it. I see this diff in ggplot2:

Screenshot 2026-03-10 at 20 59 10

MichaelChirico:
I guess I assumed the logic is choosing _among known generics_ but maybe roxygen2 doesn't have that logic b/c of how informal S3 is?

even though ggplot2 can know vctrs generics it may be harder for generics owned by the package currently being compiled unless we already scrape simple UseMethod() generics?

hadley:
Hmmmm, now that I think about it maybe the new results are correct: `vctrs::vec_cast.character` is the generic.

PR #1786, opened 2026-03-09
Link to topics, not filenames

### LINKED ISSUE #1709: Anchor is to file name instead of an alias defined on the file
I was checking one package I'm working on and it reported a note:

```
checking Rd cross-references ... NOTE
  Non-topic package-anchored link(s) in Rd file 'bioc_cran_archived.Rd':
    ‘[tools:CRANtools]{tools::CRAN_package_db()}’
  
  Non-topic package-anchored link(s) in Rd file 'cran_doom.Rd':
    ‘[tools:CRANtools]{tools::CRAN_package_db()}’
  
  See section 'Cross-references' in the 'Writing R Extensions' manual.
```

The help file is generated with something like this:

``` r
library("roxygen2")
  #' Playing with links
  #'
  #' @seealso [tools::CRAN_package_db()]
  #' @md
  null <- function () NULL
")[[1]]

out
#> \name{null}
#> \alias{null}
#> \title{Playing with links}
#> \usage{
#> null()
#> }
#> \description{
#> Playing with links
#> }
#> \seealso{
#> \code{\link[tools:CRANtools]{tools::CRAN_package_db()}}
#> }

rdfile <- tempfile()
cat(out$format(), file = rdfile)
tools::checkRd(rdfile)
help("CRANtools", package = "tools")
#> No documentation for 'CRANtools' in specified packages and libraries:
#> you could try '??CRANtools'
```

<sup>Created on 2025-04-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The file in question has several alias but the name (CRANtools) is not one of them: CRAN_package_db, CRAN_check_results, CRAN_check_details, CRAN_check_issues, summarize_CRAN_check_status, R_CRAN_WEB, R_CRAN_SRC, CRAN_current_db, CRAN_aliases_db, CRAN_rdxrefs_db, CRAN_archive_db, CRAN_authors_db. 

It would be great if it generated a cross reference with the same anchor as the function being used.

hadley:
Fixes #1709

@gaborcsardi could you please double check this? My reading of R-exts suggests that as of R 4.1 we should provide topics, not file names.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@
   * Inline R code (`` `r expr` ``) in non-indented list continuation lines no longer causes an error (#1651).
   * Link text now supports non-code markup like bold and italic, e.g., `[*italic text*][func]` generates `\link[=func]{\emph{italic text}}`, matching R's support for markup in `\link` text in R 4.5.0.
   * Links now do a better job of resolving package names: the process is cached for better performance (#1724); it works with infix operators (e.g. `[%in%]`) (#1728); no longer changes the link text (#1662); and includes base packages when reporting ambiguous functions (#1725).
+  * Links to external packages now use the topic alias instead of the Rd file name as the anchor. This fixes "Non-topic package-anchored link(s)" notes from R CMD check (#1709).
 * Package documentation improvements:
   * Correctly handles multiple arbitrary comments in the `comment` argument of `person()` in `Authors@R` (#1746).
   * Multiple email addresses in `Authors@R` now generate separate `\email{}` tags instead of a single comma-separated one (#1689).
--- a/R/markdown-link.R
+++ b/R/markdown-link.R
@@ -121,12 +121,11 @@ parse_link <- function(destination, contents, state) {
 
   ## if (is_code) then we'll need \\code
   ## `pkg` is package or NA
-  ## `fun` is fun() or obj (fun is with parens)
-  ## `is_fun` is TRUE for fun(), FALSE for obj
-  ## `obj` is fun or obj (fun is without parens)
+  ## `fun` is fun() or topic (fun is with parens)
+  ## `is_fun` is TRUE for fun(), FALSE for topic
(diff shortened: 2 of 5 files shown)


llrs:
I think you are referring to this piece of [R-exts](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Cross_002dreferences):

> There are two other forms with an optional ‘anchor’ argument, specified as `\link[pkg]{foo}` and `\link[pkg:bar]{foo}`, to link to topics _foo_ and _bar_ respectively in the package _**pkg**_. They are currently only used in HTML help (and ignored for hyperlinks in LaTeX conversions of help pages). One should be careful about topics containing special characters (such as arithmetic operators) as they may result in unresolvable links, and preferably use a safer alias in the same help page.
> 
> Historically (before R version 4.1.0), links of the form `\link[pkg]{foo}` and `\link[pkg:bar]{foo}` used to be interpreted as links to files _`foo.html`_ and _`bar.html`_ in package **_pkg_**, respectively. For this reason, the HTML help system looks for file _`foo.html`_ in package pkg if it does not find topic `foo`, and then searches for the topic in other installed packages. To test that links work both with both old and new systems, the pre-4.1.0 behaviour can be restored by setting the environment variable _R_HELP_LINKS_TO_TOPICS_=false. 

Note also the message on [NEWS](https://stat.ethz.ch/R-manual/R-devel/doc/html/NEWS.html):

> - For HTML help (both dynamic and static), Rd file links to help pages in external packages are now treated as references to topics rather than file names, and fall back to a file link only if the topic is not found in the target package. The earlier rule which prioritized file names over topics can be restored by setting the environment variable _R_HELP_LINKS_TO_TOPICS_ to a false value. 

My reading of these lead to the same conclusion [as you](https://llrs.dev/post/2024/12/23/cran-files-3/). I doubt anyone set `_R_HELP_LINKS_TO_TOPICS_=FALSE`

PR #1793, opened 2026-03-10
Dedent before processing markdown

### LINKED ISSUE #1102: Issue with bullet lists in `@param`
Surfaced in r-lib/vctrs#1100.

This block:

```r
#' Foo
#' @name foo
#'
#' @param foo
#'   * A
#'   * B
#'   * C
NULL
```

produces a nested bullet list:

```Rd
\arguments{
\item{foo}{\itemize{
\item A
\itemize{
\item B
\item C
}
}}
}
```

The first workaround is to remove the indenting:

```r
#' Bar
#' @name bar
#'
#' @param bar
#' * A
#' * B
#' * C
NULL
```

```Rd
\arguments{
\item{bar}{\itemize{
\item A
\item B
\item C
}}
}
```

Or to add a header:

```r
#' Baz
#' @name baz
#'
#' @param baz
#'   Header.
#'   * A
#'   * B
#'   * C
NULL
```

```Rd
\arguments{
\item{baz}{Header.
\itemize{
\item A
\item B
\item C
}}
}
```

hadley:
Fixes #1102

@gaborcsardi this seems "obviously correct" to me, so can you please double check my thinking?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 * Documenting data objects now generates `\usage{data(mydata)}` when the package doesn't have `LazyData: true` in its DESCRIPTION (#1425).
 * Fixed a performance regression where `roxygenize()` was very slow when the package contained large non-function objects like datasets (#1720).
 * Markdown improvements:
+  * Indented bullet lists in `@param` and other two-part tags are no longer incorrectly nested (#1102).
   * Horizontal rules (e.g. `----`) now generate a clear warning instead of an internal error about an unknown `thematic_break` xml node (#1707).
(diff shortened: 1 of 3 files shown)


gaborcsardi:
What if the first (and only?) line starts with enough whitespace to create a markdown code block? Isn't that a problem?

hadley:
Then that whitespace gets stripped and it's formatted as regular. I feel like this is unlikely to cause too many problems since most people use fenced code blocks these days.

gaborcsardi:
OK, let's do it then.

gaborcsardi (approved):

PR #1810, opened 2026-03-16
Refactor R6 implementation

hadley:
Basic idea is to introduce new intermediate classes to make it easier to test extraction from R6 objects vs. generation Rd code. I've also split up the code into multiple files.

I've done my best to avoid any substantive changes to the code, but I suspect some small changes will have crept in. Nevertheless, I think this change is worth it, since it puts R6 development on a much more solid footing for the future.

### DIFF
--- a/R/rd-r6-methods-inherited.R
+++ b/R/rd-r6-methods-inherited.R
@@ -21,33 +21,24 @@ format.rd_r6_inherited <- function(x, ...) {
     "><summary>Inherited methods</summary>"
   )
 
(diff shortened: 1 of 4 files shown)


hadley:
@gaborcsardi I don't think there's much point trying to review this thoroughly (especially by looking at the diffs), but it would be useful if you'd take a quick look and confirm that it makes sense to you. If, while you're reading, you see other obvious refactorings, feel free to just do them and then push back into this branch.

PR #1815, opened 2026-03-17
Tweak R6 usage

### LINKED ISSUE #1258: R6 methods shouldn't have crossreferences to base functions of the same name
The upcoming `roxygen2` 7.0 release adds new support for R6 classes, including generating doc entries for class methods. `pkgdown` is mistakenly interpreting these as any other code example and adding crossreference links to other packages' man pages. 

In the `arrow` package, we have `R6` classes that wrap file/connections in Arrow C++, and they have some methods that are the same name as `base` functions for working with connections in R. For example, `roxygen2` created this Rd chunk for a method:

```
\if{html}{\out{<hr>}}
\if{html}{\out{<a id="method-seek"></a>}}
\subsection{Method \code{seek()}}{
\subsection{Usage}{
\if{html}{\out{<div class="r">}}\preformatted{RandomAccessFile$seek(position)}\if{html}{\out{</div>}}
}
```

`pkgdown` rendered that as

```
<p><hr>
<a id="method-seek"></a></p><h3>Method <code><a href='https://rdrr.io/r/base/seek.html'>seek()</a></code></h3>

<h3>Usage</h3>
<p><div class="r"></p><pre>RandomAccessFile$seek(position)</pre><p></div></p>
```

The link to `https://rdrr.io/r/base/seek.html` shouldn't be there.

hadley:
Consistently refer to methods in the same way in every location.

Fixes #1258

### DIFF
--- a/R/rd-r6-methods-self.R
+++ b/R/rd-r6-methods-self.R
@@ -36,14 +36,14 @@ format.rd_r6_method <- function(x, ...) {
   }
 
   # Anchor and heading
+  call <- r6_method_name(x$class, x$name)
 
   id <- paste0("method-", x$class, "-", x$name)
   push(rd_if_html("<hr>"))
   push(rd_if_html('<a id="', id, '"></a>'))
   push(rd_if_latex("\\hypertarget{", id, "}{}"))
 
-  nm <- if (x$name == "initialize") x$class else x$name
-  push(paste0("\\subsection{Method \\code{", nm, "()}}{"))
+  push(paste0("\\subsection{\\code{", call, "()}}{"))
 
   # Description
   if (length(x$description) > 0) {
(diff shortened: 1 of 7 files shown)


gaborcsardi (approved):

hadley:
Here's a concrete example to consider: 

Screenshot 2026-03-18 at 12 41 00
Screenshot 2026-03-18 at 12 41 25

It seems bad that we show the method in three different ways: `Class$method()`, `method()`, and `obj$method()`. I think I'm leaning towards `Class$method()` being the least worse option, but we should use it consistently everywhere. I think that means we can probably drop "Method" from the title, since it will be obvious from the context?

hadley:
Overall, I think I like this approach, but it does feel a bit repetitive for simple methods:

Screenshot 2026-03-18 at 12 52 01

PR #1819, opened 2026-03-17
document rstudio reliance in inheritDotParams

kevinushey:
So that we don't break RStudio completions down the line.

Related: https://github.com/r-lib/roxygen2/issues/1808


hadley:
Thanks!

PR #1831, opened 2026-03-19
Add method description to method list

### LINKED ISSUE #1150: Provide more information in R6 methods list
In initial methods nav list, include more information, maybe either the first sentence of the description, or all the arguments (without default values), or both?

Also look into making this section denser by formatting it like usage.

hadley:
Fixes #1150

@gaborcsardi I don't have any strong feelings here, but I thought I'd try it out. It's nice that you can now see what a method does, but it does mean the individual sections now have less new information.


gaborcsardi:
Yeah, I think I would slightly prefer not to merge this. Usually you can kind of guess what a method does from its name, I think, and this PR makes it harder to quickly glance through the methods if there is a lot of them. But I don't have strong feelings, either.

hadley:
Given that neither of us have strong feelings, I think that's an argument in favour of the status quo.

PR #1837, opened 2026-03-20
Optimisation 2

hadley:
> Your goal is to make this package faster using the debrief package to analyze the performance identify bottlenecks. Begin by running debrief on `roxygenize(load_code = "installed")`. Identify the code paths that take the longest amount of time and brainstorm possible optimizations. Write down a plan for each one `optimization-plan.md`. 
> 
> Next, work systematically through each bottleneck, creating a microbenchmark to measure the performance before and after your proposed change. You MUST NOT change the behaviour of the code; if a tests fails the experiment is a failure. 
Record your results in `optimization-results.md`. Each experiment should be a commit: if the experiment is successful check in the code and `optimization-results.md`; if the experiment does not yield a meaningful improvement, just check in the `optimization-results.md` with the results and your analysis.


mcol:
I've tested this quite eagerly on Luminescence, and unfortunately I have to report a minor slow down compared to the current `main` branch. Measuring with ` system.time(roxygen2::roxygenise())`, user time went from 6.6s to 7s (median from 5 runs each), on a Linux laptop with Intel Ultra 7 and R 4.5.0.

PR #1852, opened 2026-04-06
Add argument selection to `@inheritParams`

hadley:
Fixes #1849

@maelle another one you might enjoy 😄


vignettes/reuse.Rmd
```diff
@@ -177,7 +177,11 @@ Since no other function in dplyr has an argument with this name, its documentati
 
 ### Recursive inheritance
 
-`@inheritParams` (like all `@inherits` functions) works recursively, so if `g` inherits parameters from `h`, then `f` can also inherit those parameters from `g`. However, this technique is best used sparingly: it's very easy to create complex dependency webs that are hard to reason about where making changing the documentation in one function cascades out in unexpected ways across your package. If you want to be more explicit, you should consider writing helper functions and using inline R code, as described below.
+`@inheritParams` (like all `@inherits` functions) works recursively, so if `g` inherits parameters from `h`, then `f` can also inherit those parameters from `g`. However, this technique is best used sparingly: it's very easy to create complex dependency webs that are hard to reason about where making changing the documentation in one function cascades out in unexpected ways across your package. You can avoid this problem by being more explicit about which parametrs are inherited:
```
maelle:
```suggestion
`@inheritParams` (like all `@inherits` functions) works recursively, so if `g` inherits parameters from `h`, then `f` can also inherit those parameters from `g`. However, this technique is best used sparingly: it's very easy to create complex dependency webs that are hard to reason about where making changing the documentation in one function cascades out in unexpected ways across your package. You can avoid this problem by being more explicit about which parameters are inherited:
```

vignettes/reuse.Rmd
```diff
@@ -177,7 +177,11 @@ Since no other function in dplyr has an argument with this name, its documentati
...
+
+-   `@inheritParams foo x y` inherits only `x` and `y`.
+-   `@inheritParams foo -z` inherits all parameters except `z`.
```
maelle, line 189: very nice

next, wild cards?

maelle (commented):
Nice! 😸

vignettes/reuse.Rmd
hadley: I think that leads you right back to the "what the heck is actually getting inherited" problem.

PR #1873, opened 2026-04-27
Add special-case S7 class names for S7::class_any and S7::class_missing

### LINKED ISSUE #1871: Complicated doc names for S7 methods dispatching on `S7::class_any`
The method documentation name and alias for `S7::class_any` is not human friendly. For example, the [Documenting S7](https://roxygen2.r-lib.org/dev/articles/rd-S7.html) vignette produces `\alias{size,structure(list(), class = "S7_any")-method}` when documenting the method as a separate Rd.

MRE:

```r
#' Size of an object
#'
#' @description
#' `size()` determines the size of an object.
#'
#' # Methods
#' `size()` is an S7 generic with methods available for the following
#' classes:
#'
#' `r doclisting::methods_list("size")`
#'
#' @param x An object.
#' @param ... Not used.
#' @returns A single number.
#' 
#' @export
size <- S7::new_generic("size", "x")

#' @inherit size
S7::method(size, S7::class_any) <- function(x, ...) {
  length(x)
}
```

A short abbreviation like `any-method` similar what is done for S7 base types (e.g. `S7::class_double` produces `\alias{size,double-method}`) would be nice.

mitchelloharawild:
Added S7 class name support for `S7::class_any` and `S7::class_missing` to `s7_class_name()`.
This avoids the base-case which deparses these types into `structure(list(), class = "S7_***")`, resulting in long and complex names and aliases.

Resolves #1871

### DIFF
--- a/tests/testthat/test-object-from-call.R
+++ b/tests/testthat/test-object-from-call.R
@@ -327,6 +327,13 @@ test_that("S7 method on S3 generic includes package prefix in class name", {
   expect_equal(obj$value$generic, "print")
 })
 
+test_that("S7 method with special classes any and missing", {
+  skip_unless_r(">= 4.3.0")
+  block <- roxy_block(tags = list(), file = "test.R", line = 1, call = quote(x))
+  expect_snapshot(s7_class_name(S7::class_any, block))
+  expect_snapshot(s7_class_name(S7::class_missing, block))
+})
+
 test_that("S7 method with unknown class type warns", {
   skip_unless_r(">= 4.3.0")
   block <- roxy_block(tags = list(), file = "test.R", line = 1, call = quote(x))
--- a/tests/testthat/test-object-from-call.R
+++ b/tests/testthat/test-object-from-call.R
@@ -330,8 +330,8 @@ test_that("S7 method on S3 generic includes package prefix in class name", {
 test_that("S7 method with special classes any and missing", {
   skip_unless_r(">= 4.3.0")
   block <- roxy_block(tags = list(), file = "test.R", line = 1, call = quote(x))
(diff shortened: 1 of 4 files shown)


tests/testthat/test-rd-usage.R
```diff
@@ -542,6 +542,28 @@ test_that("S7 union method usage shows member classes", {
...
+  expect_equal(
+    call_to_usage({
+      speak <- S7::new_generic("speak", "x")
+      S7::method(speak, S7::class_any) <- function(x) "hi"
+    }),
+    "## S7 method for class <any>\nspeak(x)"
+  )
+})
+
+test_that("S7 class_missing method usage shows missing class", {
```
hadley: Can you please combine into one test? And probably just fine to test `s7_class_name()` here.

mitchelloharawild: Done, thanks.

tests/testthat/test-object-from-call.R
```diff
@@ -327,6 +327,13 @@ test_that("S7 method on S3 generic includes package prefix in class name", {
   expect_equal(obj$value$generic, "print")
 })
 
+test_that("S7 method with special classes any and missing", {
+  skip_unless_r(">= 4.3.0")
+  block <- roxy_block(tags = list(), file = "test.R", line = 1, call = quote(x))
+  expect_snapshot(s7_class_name(S7::class_any, block))
```
hadley: `expect_equal()`?

mitchelloharawild: Done. I used `expect_snapshot()` for consistency with the following test (i.e. result with no message).

hadley:
Thanks!

PR #1882, opened 2026-05-20
Update lazy_data assignment to handle 'true' and 'yes'

### LINKED ISSUE #1881: The automatic usage for a data object only checks for `LazyData: true` in the DESCRIPTION, but not for `LazyData: yes` as allowed by Writing R extensions
Citing Section 1.1.1 of WRE, describing the format of the DESCRIPTION file:

"Several optional fields take logical values: these can be specified as ‘yes’, ‘true’, ‘no’ or ‘false’: capitalized values are also accepted."

These ways to specify logical values should also be respected by the check `roxygen2`, currently it is only checked if this option is set to 'true'.

jranke:
Closes #1881

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # roxygen2 (development version)
 
+* The automatic usage for a data object that is conditional on the `LazyData` option in the `DESCRIPTION` (see below) now correctly detects all ways to specify a true value, e.g. also `yes`, `Yes` or `True` (@jranke, #1881).
+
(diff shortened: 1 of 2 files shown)


hadley:
Thank you! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thanks!

PR #1898, opened 2026-06-30
Expand `@import` directives to explicit `importFrom()`

lionel-:
Branched from #1892 

This PR expands `@import pkg` directives to explicit imports in an `importFrom(pkg, sym1, sym2, ...)` expression. This pins the set of imported symbols between rounds of `document()`, and prevents users from getting issues at load time when a package update creates a conflict. Instead, new conflicts are deferred until the next time the maintainer documents, where the conflict causes an error that forces the maintainer to resolve it with an `except =` clause.

Ideally we'd detect symbols actually used by the package with a mechanism like `findGlobals()`. Until we can do it reliably, possibly with the help of Oak and annotations, I've opted to just import every exports.


hadley:
What if we introduced a new tag for this? Like `@importAll` or maybe `@importPattern` ?

lionel-:
That would be reasonable, especially as a first step to gain practical experience with the approach, but it would not feel like we're creating a pit of success for our users to fall into :)
I.e. the safe option would not be the default one.

hadley:
Right, I think we'd still want to change long term, but I worry that it's too big of a change to push through now (even in a minor release). If we used a new tag, I'm happy to merge this and encourage people to use it. (It would also be an interesting place to try and claude code automated patching system)

lionel-:
@hadley I moved the expansion to a new `@importAllFrom` tag.

Unlike `@import` where you can pass arbitrary arguments that get included verbatim in the `import()` call (e.g. `except = c(foo, bar)`), we need to actually parse exclusions since we're applying them.

I went for an exclusion syntax that is consistent with the existing `@inheritDotParams foo -x -y`.

```r
#' @importAllFrom dplyr -filter -lag
#' @importAllFrom rlang -":="
```

We need the imported package to be installed to figure out the exports. Other features in roxygen2 like `inherits` degrade to a no-op or fallback when a required package is not installed. Here I preferred to throw a hard error. We also warn when an excluded symbol is not exported.

R/namespace.R
```diff
@@ -158,19 +181,82 @@ block_directives <- function(blocks, env) {
...
+  by_sym <- split(imported, imported$sym)
+  conflicts <- keep(
+    by_sym,
+    \(x) length(unique(x$pkg)) > 1 && any(x$expanded)
+  )
+
+  # Re-exports aren't real conflicts: when several packages export the same
+  # object (e.g. `%>%`), importing it from more than one is harmless.
+  conflicts <- discard(conflicts, \(x) is_reexport(x$sym[[1]], unique(x$pkg)))
+
```
hadley:
```suggestion
```

R/namespace.R
```diff
@@ -158,19 +181,82 @@ block_directives <- function(blocks, env) {
...
-import_from <- function(package, funs) {
-  structure(list(package = package, funs = funs), class = "import_from")
+import_from <- function(package, funs, expanded = FALSE) {
+  structure(
+    list(package = package, funs = funs, expanded = expanded),
+    class = "import_from"
+  )
+}
+
+# Conflicting `@importAllFrom` directives are detected at document-time. An
```
hadley: Not just `@importAllFrom` right, but any `@importFrom` too?

R/namespace.R
```diff
@@ -158,19 +181,82 @@ block_directives <- function(blocks, env) {
...
+
+  bullets <- map_chr(conflicts, function(x) {
+    where <- sort_c(unique(x$pkg))
+    cli::format_inline("{.code {x$sym[[1]]}} is exported by {.package {where}}")
+  })
+
+  conflict <- conflicts[[1]]
+  example_sym <- auto_quote(conflict$sym[[1]])
+  example_pkg <- conflict$pkg[conflict$expanded][[1]]
+
```
hadley:
```suggestion
```

R/namespace.R
```diff
@@ -456,6 +551,77 @@ one_per_line_ignore_current <- function(name, x) {
...
+# object currently exported by `pkg`. This early expansion pins the set of
+# imports at document-time and prevents user-visible conflicts at load-time when
+# a package update introduces a conflict with other imported symbols.
+expand_import <- function(x) {
+  current <- peek_roxygen_pkg()
+  spec <- parse_import_all_from(x$val)
+  pkg <- spec$pkg
+  excluded <- spec$excluded
+
+  # Ignore an `@importAllFrom` for the package being documented
```
hadley, line 576: Why?

R/namespace.R
```diff
@@ -456,6 +551,77 @@ one_per_line_ignore_current <- function(name, x) {
...
+    character()
+  } else {
+    import_from(pkg, exports, expanded = TRUE)
+  }
+}
+
+# Splits an `@importAllFrom` value into the package to expand and the symbols to
+# leave out. An exclusion is a word with a `-` prefix, e.g. `-abort`. Exclusions
+# may be quoted: `-"-.Date"`, ``-`-.Date` ``, or `-'-.Date'`.
+parse_import_all_from <- function(vals) {
```
hadley: Would it be better to use `select_args_text`?

tests/testthat/test-namespace.R
```diff
@@ -434,6 +434,240 @@ test_that("import doesn't quote if comma present", {
...
+  out <- roc_proc_text(
+    namespace_roclet(),
+    "
+    #' @import utils
+    NULL"
+  )
+
+  expect_equal(out, "import(utils)")
+})
+
```
hadley:
```suggestion
```

R/namespace.R
lionel-: This is taken from existing code, which looks like comes from: https://github.com/r-lib/roxygen2/commit/9386fd24775ff838a6dfc860f2892d1709223037

R/namespace.R
lionel-: See last commit:

- Now parses tags with `tag_two_part()`
- Then parses arguments with `select_args_text()`
- New `class` argument threaded through `roxy_warn` and used to handle failures

R/namespace.R
```diff
@@ -33,6 +33,24 @@
 #' # This results in the following lines in `NAMESPACE`:
 #' # importFrom(magrittr,"%>%")
 #' # import(rlang)
+#'
+#' # There is a new experimental way to bulk-import a package:
+#' #' @importAllFrom rlang
+#'
+#' # This results in the following lines:
+#' # importFrom(rlang,
```
hadley, line 41: It occurs to me that the downside of this approach is that now if rlang removes a symbol, this package will fail to load. With `@import` (as long as you didn't actually use the symbol) that isn't a problem.

lionel-: oooh that's a very good point. This pretty much kills the approach.

The alternative is that once Oak symbol resolution is robust enough, we wrap it in an R package and then use that the way you'd use {globals} to figure out from usage the minimum set of symbols to import.

PR #1899, opened 2026-07-01
Use `@returns` in getting started vignette

peterdesmet:
The [documentation](https://roxygen2.r-lib.org/reference/tags-rd-functions.html) states that `@return` is superseded by `@returns`. `@returns` is indeed consistently used throughout, except in this getting started vignette.


hadley:
Thanks!

PR #1901, opened 2026-07-05
Replace markdown escaping with a combined md/Rd tokenizer

hadley:
Stacked on #1900.

This replaces `escape_rd_for_md()`/`unescape_rd_for_md()` and the double-escaping pass with an explicit grammar for how Rd and markdown mix in roxygen comments. The two languages only overlap at the backslash, so a single byte-wise C++ scan (`src/tokenizeMd.cpp`) classifies every backslash-initiated construct:

```
bs_token := "\" NAME braces*   NAME in verbatim list -> tag + all brace args
          | "\" NAME           other tags: args remain markdown
          | "\" PUNCT          two-character escape, e.g. \% \\ \{
          | "\"                lone backslash (never consumes a backtick)
```

Each token becomes an inert private-use-area placeholder, so commonmark parses text containing **no backslashes at all** — no escaping, no un-escaping, no placeholder-collision handling. Tokens are restored during the XML→Rd walk with the escaping appropriate to where they landed: as-typed in regular text, Rd-escaped inside `\verb{}`/`\code{}`/`\preformatted{}`, raw inside generated `\Sexpr{}` bodies. `\[` and `\]` pass through untokenized, because they're commonmark bracket escapes — that single rule is what used to require the de-duplication special case in `double_escape_md()`.

Highlights:

* The first commit adds characterization tests pinning current behaviour for the tricky interactions (Rd tags inside code spans/blocks, links, tables, HTML, multi-line arguments, backslash escapes) before any machinery changed.
* Rd-tags-inside-indented-code-blocks need no block-level parsing at all: the placeholder lands in a verbatim context and is restored as typed, matching the old behaviour by construction.
* `escape_verb()` loses its hidden ordering dependency on `double_escape_md()`; all escaping is now local and explicit.
* Re-documenting roxygen2 itself with the new pipeline leaves `man/` byte-identical.
* ~2.3x faster on tag-heavy documents (25KB, 400 Rd tags), ~1.3x on typical small texts; profiling shows remaining time is the commonmark parse and XML walk, i.e. the escaping overhead is gone entirely.
* One deliberate behaviour change (NEWS): `\%` in markdown text now produces `\%` in the Rd file instead of `\\%`, which rendered as a backslash followed by a comment that truncated the line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hadley:
Follow-up commit fixes the two pre-existing warts the characterization tests documented:

* `[bar\]` (escaped closing bracket) no longer leaks a link reference definition into the Rd output — reference content must not end with a backslash.
* An Rd tag whose brace argument never closes (e.g. `\code{a % b}`, where the unescaped `%` comments out the closing brace) now warns with a hint to write `\%`, instead of being silently reinterpreted as markdown. Output is unchanged.

Both have NEWS bullets; `man/` remains byte-identical when self-documenting.

hadley:
@gaborcsardi very low priority but this is take two at re-doing the escaping/unescaping. I think it's a lot cleaner now.
It doesn't make much difference performance wise, except if you have large sections with tons of Rd tags.

PR #1902, opened 2026-07-05
Resolve link topics against a per-run alias index

hadley:
Profiling `roxygenise()` on a link-heavy package (testthat) showed ~50% of the roxygen work inside link/topic resolution: `has_topic()` answers every query with a `utils::help()` call (~1.5ms), and `find_package_lookup()` maps it over all dependencies plus base packages, so each unique linked topic costs ~50 help-index searches. The per-topic cache added in #1724 only deduplicates repeated topics — every miss still pays the full scan.

This PR replaces the `help()` calls with a cached topic index per package:

* For installed packages, read the `help/aliases.rds` index that ships with the package (once per package per session — installed packages don't change between runs).
* For packages loaded from source — the package being documented, or a dependency loaded with `pkgload::load_all()` — reuse `pkgload::dev_topic_index()`, the alias index that `dev_help()` already maintains. (pkgload cached this before too, but the old path reached it through `help()` → shim → `dev_topic_parse()`/`normalizePath()` on every call, and the cache never applied to the ~50 installed dependencies, which were the bigger cost.)
* Topics are stored as a hashed environment so each `has_topic()` is a single `env_has()`; `%in%` on the alias vector rebuilt a hash table per membership test, which the R6 and Rd-inheritance paths (no per-topic cache in front of them) paid heavily.
* `pkg_deps()` is also cached per run — it re-read and re-parsed `DESCRIPTION` for every unresolved topic.
* At the start of each run, `find_package_cache_reset()` evicts only the documented package's entry (from our cache and pkgload's) — its topics are the only ones that change between runs. Failed lookups are never cached, so installing a missing dependency mid-session is picked up immediately.

`check_topic()` (validation of explicitly-qualified links) and the Rd-inheritance and R6 code paths that use `has_topic()` get the same speedup for free. Re-export attribution (`find_source()`), the ambiguous-topic warning, and all warning messages are unchanged.

Benchmark on a quiet machine (medians of 4 adjacent steady-state runs, spreads ±0.04s): `roxygenise()` on testthat goes from **3.00s to 2.15s (1.4x)**, and to **2.05s (1.5x)** combined with the markdown tokenizer rewrite in #1901. Note ~0.9s of every run is fixed `load_all()` + Rd-writing overhead, so the roxygen-work portion roughly halves. Link resolution falls from ~50% of the profile to under 8%.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # roxygen2 (development version)
 
+* Markdown link targets are now resolved against a per-run index of each package's help topics, instead of one `help()` call per topic and package. This substantially speeds up documenting packages with many cross-reference links, e.g. `roxygenize()` on testthat is about a third faster.
 * S7 methods for `[`, `[[`, `[<-`, and `[[<-` now generate valid usage (#1883).
 * `Config/roxygen2/` flag fields in `DESCRIPTION` (like `markdown`) are now parsed case-insensitively, so `true` and `True` work as well as `TRUE`, and an invalid value gives a clear error (#1875).
 * `@section` titles can now contain code that includes a colon (#1878).
--- a/R/markdown-link-resolve.R
+++ b/R/markdown-link-resolve.R
@@ -40,6 +40,7 @@ find_package_cache <- new_environment()
 # run in roxygenize() because the documented functions might change between runs
 find_package_cache_reset <- function() {
   env_unbind(find_package_cache, env_names(find_package_cache))
+  env_unbind(pkg_deps_cache, env_names(pkg_deps_cache))
 
   # Only the documented package's topics change from run to run, so evict
   # it from the topic cache (ours and pkgload's) and keep everything else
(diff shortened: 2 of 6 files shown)


hadley:
Added the last two low-hanging items the combined-branch profile surfaced (8.2% + 5.0% of samples):

* `pkg_deps()` re-read and re-parsed `DESCRIPTION` for every unresolved topic — now cached per package dir, reset each run.
* `has_topic()` used `%in%` on the alias vector, rebuilding a hash table per membership test — `pkg_topics()` now caches a hashed environment, so each lookup is a single `env_has()`. This mostly benefits the R6 and Rd-inheritance paths, which call `has_topic()` directly without the link path's per-topic cache.

With these, link/topic resolution is essentially free; the remaining profile is the markdown XML walk (~23%, a structural project) and diffuse Rd assembly.

PR #1917, opened 2026-09-04
stats is Imports

MichaelChirico:
https://github.com/r-lib/roxygen2/blob/13e78928a4e4aba09eda0dd3206a4a256c33664d/NAMESPACE#L297

This one at least gives me pause because I hope eventually this become `base::setNames` 😅

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,6 +32,7 @@ Imports:
     R6 (>= 2.1.2),
     rdtools (>= 0.1.0),
     rlang (>= 1.1.0),
+    stats,
     utils,
     withr,
     xml2


hadley:
Weird that this isn't a check error!

MichaelChirico:
I think because there are 1000s of violations 😅

hadley:
💀 TIME FOR ANOTHER GREAT REAPING OF CRAN PACKAGES 💀

MichaelChirico:
Realistically I could imagine this playing out like the thing with "Lost braces" in Rd files:

https://github.com/r-devel/r-dev-day/issues/110

